PR #505, opened 2026-06-17
feat(api): share item-name backfill endpoint

FSM1:
## Summary

Closes the last open item from **Phase 48 (REQ-4)**: encrypting share `itemName` at rest now works for legacy shares too.

Phase 48 added at-rest `itemNameEncrypted` on share **create** paths, but left a gap — there was no update endpoint accepting `itemNameEncrypted`, so the web client's lazy backfill (decision A2) could re-wrap a legacy plaintext `itemName` for the recipient but had nowhere to persist it (`void itemNameEncrypted`). This PR adds that endpoint and wires the backfill through it.

The server stays zero-knowledge: it never encrypts, only persists client-supplied ciphertext.

### `feat(api)`: share item-name backfill endpoint

- New `PATCH /shares/:shareId/item-name` — sharer-only; stores the client-supplied ECIES ciphertext as-is (mirrors the existing `PATCH /shares/:shareId/encrypted-key` pattern).
- New `UpdateItemNameDto` (even-length hex, `MaxLength` 2500) + `SharesService.updateShareItemName` (load → 404 → sharer-only 403 → persist).
- Regenerated the typed API client (`@cipherbox/api-client`) → `sharesControllerUpdateShareItemName`.
- Web: `backfillSentShareItemNames` now persists via the new endpoint, and the backfill pass is re-enabled in `fetchAllSentShares` (best-effort, idempotent via `shouldBackfill`, fire-and-forget so list loads never block).

### `docs(planning)`: bookkeeping

- Backfilled the 5 missing Phase 47 `SUMMARY.md` files (Phase 47 shipped via `#494` but its summaries were never written, which tripped a false-positive resume gate).
- Added `48-VERIFICATION.md` and advanced `STATE.md` — milestone v1.1 now 33/33 phases, 146/146 plans.

## Requirements Addressed

- **REQ-4** (Phase 48) — encrypt share `itemName` at rest, now complete for legacy rows.

## Verification

- [x] API: `apps/api` shares suites — **171 passed** (added 3 service + 3 controller specs for the new endpoint).
- [x] Web: share suites — **48 passed**; `tsc --noEmit` clean; eslint clean.
- [x] `pnpm api:generate` — generated client + `openapi.json` regenerated and committed (satisfies the `check-api-client.sh` pre-commit gate).
- [x] **E2E (two-party):** `tests/web-e2e/tests/share-itemname-backfill.spec.ts` — seeds a legacy share (plaintext `itemName`, NULL `itemNameEncrypted`), triggers the owner-side backfill via a file upload, then asserts the persisted `itemNameEncrypted` is valid ciphertext (null → hex ≠ plaintext, longer than plaintext), the recipient still displays the correct name, and the backfill is idempotent. **All 7 steps pass** against the local full stack (~52s). Runs in CI via `web-e2e.yml` (`playwright test`, unfiltered) on main-push or when dispatched against the branch.
- [ ] **Residual (out of harness):** a literal raw-`bytea` Postgres read to byte-prove "server never saw plaintext" — web-e2e has no DB client, so the spec's `GET /shares/sent` hex-≠-plaintext assertion is the in-harness proxy. Closeable later by adding a `pg` util.

## Key Decisions

- Dedicated single-field PATCH endpoint, matching the existing `/encrypted-key`, `/permission`, `/hide` convention (no generic bulk-update endpoint exists).
- Lazy client-side backfill (decision A2) rather than a server migration — the server is zero-knowledge and cannot encrypt; only the sharer holds the recipient pubkey to re-wrap.
- Sequential fire-and-forget backfill (not parallel) to stay within the API throttler; steady-state cost is ~zero via the `shouldBackfill` idempotence short-circuit.

## Summary by CodeRabbit

* **New Features**
  * Added a new `PATCH /shares/:shareId/item-name` endpoint to persist encrypted share item names (ciphertext) for legacy shares.
* **Bug Fixes**
  * Re-enabled lazy backfill for legacy shares and made it run asynchronously so share list loading doesn’t block.
  * Fixed the SDK self-bootstrap regression affecting shared folder/metadata behavior.
* **Tests**
  * Added an end-to-end Playwright test covering item-name backfill, validation, and idempotency.
* **Chores**
  * Marked Phase 47 and Phase 48 as shipped and updated tracked milestone status.
  * Updated release versions for API, web, and SDK packages.


FSM1:
Addressed the CodeRabbit review finding in `dbd65ac59`:

- **`apps/web/src/services/share.service.ts` (backfillSentShareItemNames)** — the transient plaintext `itemName` bytes from `TextEncoder().encode()` were not zeroed after wrapping. Wrapped the wrap+persist in `try/finally` and `fill(0)` the plaintext buffer (and the ciphertext buffer) after use, per the project rule to clear sensitive data from memory.

Verified: `tsc --noEmit` clean, eslint clean, web share suites 48/48 pass. Behavior-preserving, so the new e2e (`share-itemname-backfill.spec.ts`) is unaffected.

This was CodeRabbit's only actionable item (the rest of the review was file-list/info); it was an 'outside diff range' comment with no resolvable inline thread.

PR #520, opened 2026-06-18
fix(release): break sdk changelog self-compare loop

FSM1:
## Summary

`packages/sdk` was stuck in a release-please loop: its CHANGELOG kept gaining duplicate, self-comparing `0.36.0` entries on every push to `main`.

## Root cause

`release-please-config.json` pinned `packages/sdk` to `"release-as": "0.36.0"` — equal to both the manifest version and the latest tag (`@cipherbox/sdk-v0.36.0`, cut in `#508`). After `#509` (`feat(web)`) landed releasable changes under `packages/sdk/src/**` (`client.ts`, `share/*`), release-please attributed that feat to the sdk component **by path** and tried to release it — but the pin forced the version back to `0.36.0`. That produced a degenerate `compare/@cipherbox/sdk-v0.36.0...@cipherbox/sdk-v0.36.0` changelog header, and because the version never advanced (no new tag, pin never cleared), the same block was re-appended every push (`#510`, then a byte-identical dupe in `#514`).

The `release-as` pins are injected by `.github/scripts/pr-release-preview.js`; for `#509` it bumped api/crypto/api-client/sdk-core but **not** sdk, leaving the pin behind while the source changed.

## Changes

- **`release-please-config.json`** — bump `packages/sdk` `release-as` `0.36.0` → `0.37.0` so the pending `#509` feat releases as a real minor bump (`bump-minor-pre-major`), advancing the version and ending the loop.
- **`packages/sdk/CHANGELOG.md`** — remove the two duplicate self-comparing `0.36.0` blocks, leaving the legitimate `0.35.0...0.36.0` (`#507`) entry. release-please will regenerate a clean `0.37.0` section (including `#509`) on its next run.

Tags, manifest, and `package.json` were already correct and in sync — only the changelog was corrupted.

## Scope note

This fixes sdk only. The same latent trap exists for every package whose `release-as` currently equals its manifest version (api, web, core, crypto, api-client, sdk-core, plus crates). The systemic fix — hardening `pr-release-preview.js` to clear satisfied/inherited pins and reconcile its path attribution with release-please — is tracked separately as a GSD todo.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Corrected and reorganized release notes for SDK version information.

* **Chores**
  * Updated release version configuration for web and SDK packages.

### DIFF
--- a/packages/sdk/CHANGELOG.md
+++ b/packages/sdk/CHANGELOG.md
@@ -1,75 +1,53 @@
 # Changelog
 
-## [0.36.0](https://github.com/FSM1/cipher-box/compare/@cipherbox/sdk-v0.36.0...@cipherbox/sdk-v0.36.0) (2026-06-18)
-
-
-### Features
-
-* **web:** shared-folder intra-share move and useFolderNavigation consolidation ([#509](https://github.com/FSM1/cipher-box/issues/509)) ([c36ac6d](https://github.com/FSM1/cipher-box/commit/c36ac6d7792947a734a539a23de6b42d5c1fdd98))
-
-## [0.36.0](https://github.com/FSM1/cipher-box/compare/@cipherbox/sdk-v0.36.0...@cipherbox/sdk-v0.36.0) (2026-06-18)
-
-
-### Features
-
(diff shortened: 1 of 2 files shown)


FSM1:
Superseded by the renamed `chore(ci)` branch/PR (no `fix` convention for this CI maintenance).

PR #689, opened 2026-07-21
fix(core): size the encode buffer up front so secret writes never realloc

### LINKED ISSUE #681: core: codec-wide zeroizing encode buffer for encode-realloc content-key hygiene
## Context

The sealed-body content-key hygiene story is now handled at two of three layers:

- The caller's owned key material (`Version.content_key: SecretBytes`) zeroizes on drop.
- The transient `Value::Bytes` copy built by `encode_read_body` is scrubbed before the `Value` tree drops (closed by #672 / PR #680 via `Value::zeroize_bytes`).
- The final encoded plaintext buffer is zeroized by the terminal owner (`seal_read_body`).

## Gap

`codec::encode` builds its output `Vec<u8>` from empty and grows it by appending. When the `Vec` exceeds its capacity it **reallocates**: a new, larger backing buffer is allocated, the existing bytes are copied over, and the old backing buffer is **freed without being zeroized**. By the time a reallocation happens, the partially-serialized plaintext already written into the old buffer can include content-key bytes — so a copy of key material is left in freed-but-uncleared heap.

The final buffer is handled (terminal owner zeroizes it), and the intermediate `Value::Bytes` copy is handled (#680), but the **intermediate reallocated backing stores produced during `encode` growth are not**.

This is a **codec-wide, pre-existing** property, distinct from the specifically-named gap in #672: it affects every `encode` of secret material — content keys today, and grant / owner / history seed material as those structures land. #672's two suggested closure directions (a scrubbed `Value::Bytes` copy, or a zeroizing `Value` variant) do **not** address the reallocation path; only a zeroizing encode buffer does.

## Suggested direction

A zeroizing encode path for secret-carrying structures, e.g. one of:

- Encode into a `Zeroizing<Vec<u8>>` accumulator whose backing store is wiped on any reallocation (custom growth, or pre-reserve exact capacity so no realloc occurs mid-encode).
- A two-pass encode: compute the serialized length first, `reserve_exact` once, then encode — guaranteeing zero reallocations during the secret write, so only the single terminal buffer holds key bytes.
- A secret-aware encoder variant used specifically on the sealed-body (and later grant/owner/history) paths.

No wire-format impact — the serialized bytes are unchanged; this is purely allocation hygiene.

## Provenance

Raised during implementation of #672 (PR #680, which closed the named `Value::Bytes`-copy gap). Deliberately not fixed there to keep that slice scoped to the specifically-flagged transient copy; filing here so the codec-wide reallocation-hygiene hardening is tracked. Same follow-up pattern as #672 itself (spun out of the #620 seal-slice review).

FSM1:
## What landed

`codec::encode` built its output `Vec<u8>` from empty and grew it by appending. On capacity overflow the `Vec` reallocates: it allocates a larger backing store, copies the bytes, and frees the old one **without zeroizing**. Any secret plaintext already serialized into that old buffer — inline `contentKey` bytes today, grant / override / history seed material as those structures land — was thus stranded in freed-but-uncleared heap.

This makes `encode` **two-pass**: a new `encoded_len(value)` computes the exact serialized byte count, `encode` reserves that once with `Vec::with_capacity`, then writes. With a single right-sized allocation the write never reallocates, so the one terminal buffer — which its owner already zeroizes (`seal_read_body`, the grant/override/history seal paths) — is the only allocation that ever holds key bytes.

Applied at the single `codec::encode` entry point, so it covers every secret-carrying encode path codec-wide (`encode_read_body`, `encode_grant_blob_payload`, `encode_override_seed_payload`, `encode_history_link_payload`) uniformly — no per-path variant.

## No wire-format impact (KAT byte-identity)

Only the buffer's initial capacity changed; the emitted bytes are unchanged. Verified by regenerating the full KAT corpus (`cargo run -p cipherbox-core --example kat_gen`) and confirming a **byte-identical** `git diff` on `crates/core/kat` (empty). The native + WASM KAT suites pass unchanged.

## Realloc-free proof (tests added)

- `secret_bearing_encode_never_reallocates` (inline unit, `src/codec/encode.rs`) — encodes a file-body-shaped value carrying 32-byte `contentKey` entries, sized to force multiple growths from empty. Captures the backing pointer and capacity after the reserve and asserts **both are unchanged** after the write: a reallocation would move the buffer and/or grow capacity, so this proves the single-allocation invariant independent of allocator over-allocation.
- `encoded_len_matches_emitted_bytes` (proptest, `tests/codec_props.rs`) — asserts `encoded_len(v) == encode(v).len()` for arbitrary values across the full profile data model. This is the lockstep invariant that keeps the length oracle exact (and thus the reservation short-proof-free) as the encoder evolves.
- `encoded_len_is_exact` (inline unit) — pins exactness across every head size class and nesting.

## CI gate wiring

No new required check. The tests land in the crate's existing gates:
- `encoded_len_matches_emitted_bytes` runs in **Core KATs (native + WASM)** on all three legs (native, `wasm32-wasip1`, and the browser-shaped `wasm32-unknown-unknown` leg, which explicitly names `codec_props`), and in **cargo-linux**'s workspace test.
- The inline unit tests run in the native `cargo test -p cipherbox-core` leg and the `wasm32-wasip1` leg.

`cargo fmt`, `cargo clippy -D warnings`, and the full native core suite are green; `wasm32-wasip1` compiles cleanly (pointer-width parity).

Closes #681

## Summary by CodeRabbit

* **New Features**
  * Added a public `encoded_len` helper to report the exact encoded size of values.
  * Improved encoding efficiency by preallocating the required output space.

* **Bug Fixes**
  * Ensured encoded length calculations consistently match the produced output.

* **Tests**
  * Expanded coverage across nested and varied value types.
  * Added checks preventing unnecessary buffer reallocation during encoding.

### DIFF
--- a/crates/core/src/codec/encode.rs
+++ b/crates/core/src/codec/encode.rs
@@ -18,13 +18,59 @@ pub(super) const SIMPLE_NULL: u8 = 22;
 
 /// Encode a value in the deterministic profile.
 ///
+/// Two-pass: [`encoded_len`] sizes the buffer up front so the single allocation
+/// is exact and the write never reallocates. A grow-from-empty `Vec` frees each
+/// intermediate backing store un-zeroized, stranding secret bytes (content keys,
+/// scope/override/history seeds) in freed heap; with one right-sized buffer only
+/// the terminal owner's plaintext ever holds them, and that owner zeroizes it.
+///
 /// Callers must respect [`super::MAX_DEPTH`]: a deeper `Value` still encodes
 /// (encoding is infallible), but its bytes reject as `depth-exceeded` on
 /// decode — debug builds assert the bound to catch that divergence early.
 pub fn encode(value: &Value) -> Vec<u8> {
-    let mut out = Vec::new();
+    let mut out = Vec::with_capacity(encoded_len(value));
     write_value(&mut out, value, 0);
     out
+}
+
+/// The exact number of bytes [`encode`] emits for `value`. Mirrors
+/// [`write_value`] head-for-head; the codec property suite pins the two in
+/// lockstep (`encoded_len == encode().len()`), which is what makes [`encode`]'s
+/// single up-front reservation provably realloc-free.
+pub fn encoded_len(value: &Value) -> usize {
+    match value {
+        Value::Unsigned(n) | Value::Negative(n) => head_len(*n),
+        Value::Bytes(b) => head_len(b.len() as u64) + b.len(),
+        Value::Text(t) => text_len(t),
+        Value::Array(items) => {
+            head_len(items.len() as u64) + items.iter().map(encoded_len).sum::<usize>()
+        }
+        Value::Map(map) => {
+            head_len(map.len() as u64)
+                + map
+                    .entries()
+                    .iter()
+                    .map(|(k, v)| text_len(k) + encoded_len(v))
+                    .sum::<usize>()
+        }
+        Value::Bool(_) | Value::Null => 1,
+    }
+}
+
+/// The byte length of a text item: its head plus its UTF-8 bytes.
+fn text_len(t: &str) -> usize {
+    head_len(t.len() as u64) + t.len()
+}
+
+/// The byte length of a shortest-form head for `arg` (mirrors [`write_head`]).
+fn head_len(arg: u64) -> usize {
+    match arg {
+        0..=23 => 1,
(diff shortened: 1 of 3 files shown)


FSM1:
@coderabbitai full review

FSM1:
@coderabbitai full review please

Issue #628, opened 2026-07-20
engine: build the gated resolve and publish pipeline and liveness jobs


FSM1:
## Scope

The net module: cache-first resolve (last-known-good renders immediately, fan-out GET across the endpoint set, core verify, adoption gate on every resolve); publish register-first fail-closed through the API client, core-signed with sequence 1 embedded on first publish and exact expected sequence on CAS, parallel PUT with any-ack success and background retry, confirm-by-re-resolve, lost-race re-resolve; explicit TTL from the sync timing profile + the 90-day client-signed EOL; the ~hourly keyless re-PUT Scheduler job, the sub-~30-day EOL republish at seq+1, revival after a >EOL lapse via the recovery endpoint, and retirement policy (root linger stubbed pending the migration-window constant).
Does NOT land: rebase, the pointer planes.

## Blueprint anchors

`blueprint/engine.md`: Resolve/publish pipeline; API client (register-first ordering); Open edges

## Gate

Engine simulation: CAS races, any-ack/retry fan-out, virtual-time multi-day EOL timelines, the re-PUT job, revival; contract-suite additions: register-first fail-closed, batch register/retire idempotency.

## Depends on

- #625 (engine: add the hand-written API client, tokens, and the contract suite)
- #626 (engine: implement the adoption gate and floor law with the sim harness)
- #627 (api: build the pin and name registry with register-first and quota)

Issue #626, opened 2026-07-20
engine: implement the adoption gate and floor law with the sim harness


FSM1:
## Scope

The six-stage adoption gate as a pure pipeline over core's checks: record verify, commitment verify (scope roots), grant-section authentication under committed writer pseudonyms, strictly-newer sequence vs the durable per-name floor, epoch tag at or above the scope's epoch floor, unseal success — any failure a fail-closed trust violation, never staleness; whole-record rejection, pin last-known-good. The floor law: advance only on AAD-confirmed unseal, cold-seed from a re-point object's owner-vouched writeEpoch/minReadEpoch, pointer writeEpoch advances on sight, grant-blob epoch fields advisory only, FloorStore regression fail-closed. Lands the simulation harness itself: N engine instances on one fake record store, virtual time, adversarial replay/transplant/re-sign, table-mirroring anti-vacuity meta-tests.
Does NOT land: resolve/publish wiring, live pointer resolution — records are hand-fed.

## Blueprint anchors

`blueprint/engine.md`: Adoption gate and floors; Doctrine

## Gate

Engine simulation in the cargo workspace tests PR gate: the six-stage matrix (acceptance + every reject class with its named trust-violation error) and the floor-law scenarios, hard-asserted 1:1 against the design tables.

## Depends on

- #623 (engine: scaffold seams, fakes, conformance kits, and the facade)
- #621 (core: implement the grant section, write-body, and structure signatures)
- #622 (core: implement IPNS records, the name codec, and pointer payloads)

PR #706, opened 2026-07-21
feat(engine): implement the sync core with op queue, rebase, and pointers

### LINKED ISSUE #632: engine: implement the sync core with op queue, rebase, and pointers
## Scope

The state law (last-known-good snapshot + pending-op overlay, single owner); the durable FIFO op queue of intent ops carrying base sequences, FIFO replay through the standard rebase onto gate-passing state only; all five rebase races (conditional delete, rename/rename via parent CAS, add/add auto-suffix with the strict NFC+case-fold comparator, dest-first move with presence-conditional source-remove, dual-link observed repair via the link counter); dead-letter with staged bytes preserved; budgeted offline staging behind StagingStore; the 30 s jittered focus-window tick, immediate ticks on RefreshHintSource, on-access refresh past staleness; the staleness ladder and withheld-update escalation; the pointer planes — scope pointer and re-point object handling, consult discipline polled-not-fallback, the indexed vault-pointer chain with probe-one-past, owner-plane publishes in owner sessions only, the non-circular cold-start sequence.
Does NOT land: scope-exit rotation triggering — a cross-scope relink out of a granted scope queues the trigger event only.

## Blueprint anchors

`blueprint/engine.md`: Sync core; Pointer planes; Adoption gate and floors (cold start)

## Gate

Engine simulation: the five-race table mirrored 1:1, offline queue replay, dead-letter on revoked-while-offline, staging-budget fail-fast, staleness ladder and escalation, the keyless re-PUT adversary, cold-start adopts nothing before floor seeding.

## Depends on

- #626 (engine: implement the adoption gate and floor law with the sim harness)
- #628 (engine: build the gated resolve and publish pipeline and liveness jobs)
- #622 (core: implement IPNS records, the name codec, and pointer payloads)

### LINKED ISSUE #724: engine: dag::assemble must enforce the leaf-count invariant in all builds, not only debug_assert
## Context

Follow-up from #700 (#630 content plane, merged). Greptile posted a P1 immediately after merge on `crates/engine/src/content/dag.rs:102` — https://github.com/FSM1/cipher-box/pull/700#discussion_r3623197793 — revisiting the `debug_assert_eq!` added for the CodeRabbit nitpick that asked `assemble` to self-check the `ceil(size / chunkSize)` leaf-count invariant `decode_root` enforces.

## Problem

`assemble` (public, re-exported at the crate root) validates `leaves.len() == expected_leaf_count(plaintext_len, chunk_size)` only via `debug_assert_eq!` (`dag.rs:98`), which **compiles out in release builds**. If any caller passes an inconsistent `(leaves, plaintext_len, profile)`, a **release build silently emits a root block that this crate's own `decode_root` rejects** with `InvalidManifest { reason: "link count inconsistent with size" }` (`dag.rs:168-170`). If that root is pinned and published to IPNS, the **published version is permanently unreadable** — a silent durability failure on a canonical record that surfaces only on a later read.

Not a live bug today: the sole caller (`seal_content`, `content/mod.rs`) passes consistent values. This is latent public-API hardening.

## Why valid — revisiting the debug_assert call

The nitpick fix deliberately chose `debug_assert!` on the basis that `assemble` is the produce side over trusted engine inputs (a programmer-error guard, not an untrusted-input trust boundary). That is defensible, but severity + surface tip the other way:

- The failure is **silent and permanent** (an unreadable published version), not a dev-time crash.
- `assemble` is `pub` + crate-root re-exported — "the only caller is correct" is not a durable guarantee for a public surface.
- The durability ethos (the network is canonical; clients verify records cryptographically) means never publishing content that fails your own decoder is a fail-closed-in-all-builds invariant, not a debug-only one.

## Fix

Enforce the invariant in every build:

- Elevate `assemble` to `Result<_, DagError>` and return `DagError::InvalidManifest { reason: "link count inconsistent with size" }` (reuse the existing variant, or a produce-side reason string) when `leaves.len() as u64 != expected_leaf_count(plaintext_len, profile.chunk_size() as u64)`, **before** constructing the root block.
- Propagate the error through `seal_content` (`content/mod.rs`) and the crate-root re-export.
- Keep the shared `expected_leaf_count` helper so produce/consume stay symmetric; drop the `debug_assert`.

## Test

A plain `#[test]` (so the assertion is the returned `Err`, not a compiled-out `debug_assert`) that calls `assemble` with a deliberately mismatched leaf set / `plaintext_len` and asserts it returns `Err(DagError::InvalidManifest { .. })` rather than emitting a root `decode_root` would later reject.

## Refs

- #700 / #630 (content plane); Greptile P1 https://github.com/FSM1/cipher-box/pull/700#discussion_r3623197793

Severity: MEDIUM — latent; silent durability loss if a caller mis-wires; no live path today.

FSM1:
## What

Lands the keystone **sync core** for `crates/engine` behind the frozen facade — the one stateful sync model several downstream slices gate on. New `crates/engine/src/sync/` module with eight submodules:

- **`model`** — the working tree (the state law's left operand): a gate-passing `Snapshot` of nodes + parent links carrying the monotonic dual-link counter, plus the single strict name comparator (`collation_key`) and the `name (2).ext` auto-suffix.
- **`op`** — the five intent ops (`create`/`delete`/`rename`/`relink`/`updateContent`), each carrying its base sequence, and the durable-queue codec (engine-owned opaque bytes; content rides staging keys, never the op).
- **`overlay`** — the state law: rendered = last-known-good snapshot ⊕ pending-op overlay (optimistic, single owner).
- **`rebase`** — FIFO replay onto gate-passing state only, the five per-op race rules 1:1 (conditional delete, rename/rename via parent CAS, add/add auto-suffix, dest-first move with presence-conditional source-remove, dual-link observed repair), dead-lettering, and `decode_queue` (dead-letters corrupt entries).
- **`staging`** — the budgeted offline staging policy over `StagingStore`: uploads fail fast past the profile budget, metadata ops queue unbounded, orphan-key GC.
- **`staleness`** — the ladder (fresh → reconciling → stale → offline) and the shared-scope-only withheld-update escalation, off the injected clock.
- **`pointer`** — the scope/vault pointer planes: re-point object seal/open (over core's `RepointObject`), the indexed vault-pointer walk with probe-one-past, the owner-plane write gate (owner sessions only), the polled-not-fallback consult discipline, and the cold-start floor cold-seed (reusing `gate::floor`).
- **`tick`** — the jittered focus-window tick (jitter from injected entropy), immediate ticks on `RefreshHintSource`, the focus-set computation (vault pointer + open folder + ancestor chain + open shared-scope pointers + mailbox poll), on-access refresh, and manual-refresh nocache semantics.

Small additive edits: `NodeId` gains `Ord`/serde derives (deterministic maps + op codec), `NodeKind` gains serde; `sync` wired into `lib.rs` with re-exports.

## Why

Closes #632. Implements exactly the issue's scope against `blueprint/engine.md` ("Sync core", "Pointer planes", "Adoption gate and floors") and `CONTEXT.md`. Builds on #626 (adoption gate / floor law — reused via `gate::floor`) and #628 (net pipeline). Determinism is injected throughout: entropy only via `Entropy`, time only via `Scheduler`/`UnixMillis` — no direct clock or RNG in engine logic. All crypto stays in `crates/core` (the pointer plane calls core's `seal_pointer_payload`/`open_pointer_payload` and the pointer KDF edges; the engine implements no crypto). Every gate rejection is fail-closed.

**Explicitly deferred, per the issue scope** ("does NOT land: scope-exit rotation triggering"): a cross-scope relink out of a granted scope *queues* the scope-exit rotation trigger (`ReplayReport::scope_exit_triggers`) — it never rotates. The rotation primitives land with the rotation slice.

Scoping notes:
- The sync core operates on an engine-domain tree projection; later slices assemble it from gate-passing `Adopted` read-bodies. The tree mirrors core's `ChildRef` (id, name, kind, `link_counter`).
- `collation_key` folds case (Unicode default lowercase); canonical NFC normalization is the cross-language KAT'd surface core owns (its `name_cmp` is the wire-order sibling) and graduates there — the engine uses one comparator identically at create and merge.
- The tick loop's `on_tick` is a caller-supplied async closure so the net/pointer resolve wiring composes on top; the loop's timer-vs-hint race, jitter, and focus-set are fully covered here.

## Verification

- `cargo test -p cipherbox-engine`: **192 passed, 0 failed** (123 lib unit + 15 new sync integration + 54 existing across adoption-gate/net/facade/floor/conformance).
- `crates/engine/tests/sync.rs` is the simulation harness mirroring the issue's gate 1:1: the five-race table, offline queue replay, dead-letter on revoked-while-offline (staged bytes preserved), staging-budget fail-fast, the staleness ladder + escalation, the keyless-re-PUT adversary (floor law blocks the rollback), and cold-start-adopts-nothing-before-floor-seeding.
- `cargo clippy -p cipherbox-engine --all-targets`: clean. `cargo fmt -p cipherbox-engine --check`: clean.
- Ran `/simplify`, `/security-review`, and `/crypto-privacy-review` on the diff; findings folded in (see below).

### Review findings addressed

- **/simplify** (3 parallel reviewers): added `Snapshot::link_next` (single pass; folds the `max_link_counter+1; link` idiom out of 6 sites and pulls the fresh-link-wins invariant into the model); `replay` now constructs `ReplayReport` once (dropped a wasted init clone); single-scan `observed_repair` with `Repair.remove` narrowed to losing parent ids; `HashSet` in `orphan_staging_keys`; a shared `duration_millis` helper (was inlined in `staleness` + `tick`); an `OpResolution::applied()` constructor; deleted the dead `advance_write_epoch_on_sight` wrapper (the floor law is its home) and rewired `resolve_vault_pointer` through `open_repoint`; added coverage for the previously-uncalled `scope_pointer_name`/`scope_pointer_signer` and for `content_version`.
- **/crypto-privacy-review**: clean — no crypto in the engine (all via `crates/core`: pointer KDF edges + `seal/open_pointer_payload`), seal nonce and jitter from injected entropy, time only via `UnixMillis`, login-secret-derived `SecretBytes` zeroize on drop and caller-owned buffers are never zeroed, the vault-pointer walk and cold-seed are fail-closed. Tightened one doc: `stage_op`'s upload bytes are the already-sealed content payload (sealing runs upstream), so no plaintext lands at rest.
- **/security-review**: clean — every loop bounded (`MAX_VAULT_POINTER_PROBE`, `MAX_SUFFIX_PROBE`→`SuffixExhausted`, `ancestors` cycle guard), no unguarded panics on external input, corrupt op-queue bytes dead-letter rather than crash the replay.

### Coordinator-flagged addition (same-sequence split-brain)

Per timely guidance from the net publish slice (#692): two writers each minting a **different** record at `floor + 1` both land, and IPNS higher-sequence-wins cannot converge equal sequences. Added `reconcile_head` + `HeadReconciliation`: a deterministic byte-order tiebreak over the verified record bytes picks the canonical head with no shared state; the loser rebases its local ops onto the winning sibling and re-mints at `sequence + 1`, after which ordinary higher-sequence-wins finishes convergence. Covered by three unit tests (higher → lost race, lower/identical → converged, same-sequence divergence with symmetric winner/loser).

Closes #632
Closes #724 

## Summary by CodeRabbit

* **New Features**
  * Added a new synchronization “sync core” surface with offline replay, pending-op overlays (base + queued intents), pointer resolution/sealing, focus refresh tick loop, and staleness classification.
  * Added budgeted staging (including queued/enqueued outcomes) plus orphan staging key discovery.
  * Node identifiers and created-node records now support deterministic ordering and serialization.
* **Bug Fixes**
  * Content sealing and DAG assembly are now fail-closed: invalid manifests/mismatches return errors even outside debug builds.
  * Updated sealing to surface DAG validation failures via a dedicated sealing error type.

### DIFF
--- a/crates/engine/src/content/dag.rs
+++ b/crates/engine/src/content/dag.rs
@@ -85,21 +85,22 @@ pub struct ContentDag {
 /// `contentCid`. `plaintext_len` is the pre-seal byte length (the ranged-read
 /// size bound). Deterministic: identical leaves + length + profile give a
 /// byte-identical root block and CID.
+///
+/// Fails closed when the leaf count does not match [`decode_root`]'s
+/// `ceil(size / chunkSize)`. Enforced in every build (not a release-compiled-out
+/// `debug_assert`): a mis-wired caller must never emit a root this crate's own
+/// [`decode_root`] would reject as `link count inconsistent with size`, which
+/// would leave the published version permanently unreadable.
 pub fn assemble(
     leaves: &[SealedChunk],
     plaintext_len: u64,
     profile: &ContentProfile,
-) -> ContentDag {
-    // Produce/consume symmetry: assemble emits exactly the leaf count decode_root
-    // requires, so this crate never produces a root its own decode would reject.
-    // A debug_assert (not a fallible signature) is the right level — the inputs
-    // are trusted engine values, so this guards a mis-wired caller, not untrusted
-    // input.
-    debug_assert_eq!(
-        leaves.len() as u64,
-        expected_leaf_count(plaintext_len, profile.chunk_size() as u64),
-        "assemble leaf count must match decode_root's ceil(size / chunkSize)"
-    );
+) -> Result<ContentDag, DagError> {
+    if leaves.len() as u64 != expected_leaf_count(plaintext_len, profile.chunk_size() as u64) {
+        return Err(DagError::InvalidManifest {
+            reason: "link count inconsistent with size",
+        });
+    }
     let links = leaves
         .iter()
         .map(|leaf| Value::Bytes(leaf.cid.clone()))
@@ -110,10 +111,10 @@ pub fn assemble(
     root.insert(LINKS_KEY, Value::Array(links));
     let root_block = encode(&Value::Map(root));
     let content_cid = compute_cid(DAG_ROOT_CODEC, &root_block);
-    ContentDag {
+    Ok(ContentDag {
         content_cid,
         root_block,
-    }
+    })
 }
 
 /// The manifest read back from a verified root block: the fixed chunk size, the
@@ -243,7 +244,7 @@ mod tests {
     #[test]
     fn root_content_addresses_to_its_content_cid() {
         let (leaves, profile) = framed(&(0..40u8).collect::<Vec<_>>(), 1);
-        let dag = assemble(&leaves, 40, &profile);
+        let dag = assemble(&leaves, 40, &profile).unwrap();
         assert!(verify_cid(&dag.content_cid, &dag.root_block).is_ok());
         assert_eq!(
             dag.content_cid[1], DAG_ROOT_CODEC,
@@ -254,8 +255,8 @@ mod tests {
     #[test]
     fn assembly_is_deterministic() {
         let (leaves, profile) = framed(&(0..40u8).collect::<Vec<_>>(), 7);
-        let a = assemble(&leaves, 40, &profile);
-        let b = assemble(&leaves, 40, &profile);
+        let a = assemble(&leaves, 40, &profile).unwrap();
+        let b = assemble(&leaves, 40, &profile).unwrap();
         assert_eq!(a, b, "same leaves + length => byte-identical root");
     }
 
@@ -263,7 +264,7 @@ mod tests {
     fn root_manifest_round_trips_the_ordered_leaf_cids() {
         let plaintext: Vec<u8> = (0..40u8).collect();
         let (leaves, profile) = framed(&plaintext, 3);
-        let dag = assemble(&leaves, plaintext.len() as u64, &profile);
+        let dag = assemble(&leaves, plaintext.len() as u64, &profile).unwrap();
 
         let manifest = decode_root(&dag.root_block).unwrap();
         assert_eq!(manifest.chunk_size, profile.chunk_size() as u64);
@@ -278,7 +279,7 @@ mod tests {
     #[test]
     fn leaf_cids_use_the_raw_codec_root_uses_dag_cbor() {
         let (leaves, profile) = framed(b"abc", 5);
-        let dag = assemble(&leaves, 3, &profile);
+        let dag = assemble(&leaves, 3, &profile).unwrap();
         assert_eq!(leaves[0].cid[1], CONTENT_CID_CODEC, "leaf is raw");
         assert_eq!(dag.content_cid[1], DAG_ROOT_CODEC, "root is dag-cbor");
     }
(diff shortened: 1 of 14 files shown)


FSM1:
Merged current `main` in (brings in the #700 content plane) and folded in #724: `dag::assemble` now enforces the leaf-count invariant in all builds (returns `DagError::InvalidManifest`) instead of a release-compiled-out `debug_assert`. Closes #724 on merge.

FSM1:
@coderabbitai full review please.

PR #731, opened 2026-07-22
fix(core): bind mailbox recipient into the sender signature preimage

### LINKED ISSUE #712: core/engine: mailbox sender signature is recipient-agnostic — enforce the no-implicit-addressing contract [MEDIUM]
**Severity: MEDIUM** (design-contract; spec-conformant as-is). From the retro `/crypto-privacy-review` of #676 (IPNS/pointer/mailbox payloads).

`crates/core/src/payload/mailbox.rs:60-67` (`sig_preimage`) / `:125-148` (`verify_inner`): the sender signs over `[MAILBOX_SIG_DOMAIN, v, senderIdentityPk, payload]` only — the inner signed blob is **recipient-agnostic**. A malicious recipient R1 who HPKE-opens an item can lift the inner `{payload, senderIdentityPk, senderSig}` and re-HPKE-seal it to R2; R2's `open_mailbox_payload` verifies the sender signature successfully and concludes the sender authored that payload "to them." The untrusted API alone cannot do this (it can't open R1's blob), so the attacker must be a malicious recipient — which bounds impact.

This **matches the frozen spec** (#39 D9 / `blueprint/core.md` "Pointer payloads — Mailbox"), which deliberately binds only domain/version/senderKey. This issue is to make the contract explicit and enforced at the layer that owns mailbox semantics:

- **Confirm/enforce:** no higher-layer mailbox payload semantics may treat "I received a validly-signed item" as "the sender addressed this item to me" — anything action-triggering or grant-related must carry its own recipient binding inside `payload`.
- If recipient-binding is ever desired at the core layer, it belongs in the sig preimage.
- Add the cross-recipient relay KAT (tracked with the KAT-gap issue) to freeze the current property as intended.

FSM1:
## What

Binds the recipient's X25519 encryption public key into the mailbox sender-signature preimage in `crates/core/src/payload/mailbox.rs`. The signed preimage becomes `[domain, v, recipientEncPk, senderIdentityPk, payload]` (was `[domain, v, senderIdentityPk, payload]`).

## Why

Closes #712 (MEDIUM, design-contract). The old preimage was recipient-agnostic, so a malicious recipient R1 could HPKE-open an item, lift the untouched inner `{payload, senderIdentityPk, senderSig}`, re-HPKE-seal it to R2, and R2 would verify it as if the sender had addressed R2. Binding the recipient key makes a relayed item fail `identity-signature-invalid` for the wrong recipient.

## Wire shape unchanged

Not a wire-format or caller-signature change. `seal_mailbox_payload` already takes `recipient_enc_pub`; `open_mailbox_payload` derives the recipient public key from the secret. The block/plaintext CBOR shape (`{ct,enc}` outer, `{payload,senderIdentityPk,senderSig}` inner) is identical; only the signed preimage gains a field. No engine callers change.

## KAT inversion

The accept vector `relay-to-other-recipient-still-verifies` (pre-marked in `kat_gen.rs` as a conscious break for #712) is removed; a reject vector `relay-to-other-recipient` (`identity-signature-invalid`) is added. Mailbox vector counts: accept 3 to 2, reject 5 to 6. Regenerated `mailbox_accept.json`, `mailbox_reject.json`, `manifest.json` via `cargo run -p cipherbox-core --example kat_gen`.

## Tests

Added unit test `relay_to_other_recipient_fails_identity_check`. `cargo fmt --all --check`, `cargo test --workspace`, and `cargo clippy --workspace --all-targets` all green.

## Summary by CodeRabbit

* **Security Enhancements**
  * Mailbox payload sender signatures are now explicitly bound to the intended recipient’s encryption public key, preventing successful verification when payloads are relayed to another recipient.
  * Relayed payloads that previously could be accepted for the wrong recipient now fail with an identity-signature verification error, while valid recipient-bound messages continue to verify.

* **Tests**
  * Updated and expanded mailbox KAT vectors and generator scenarios, including new deterministic “relay-to-other-recipient” reject coverage and updated accept/reject blocks.

* **Documentation**
  * Clarified recipient-bound signature preimage details in the mailbox sealing/verification description.


FSM1:
> ### Comments Outside Diff (1)
> 1. `crates/core/src/payload/mailbox.rs`, line 64-72 ([link](https://github.com/fsm1/cipher-box/blob/82b742b8335940bacf6746c551ff10d6df6259c3/crates/core/src/payload/mailbox.rs#L64-L72))
>    [P1](#) **Existing V2 Items Become Unreadable**
>    When an old v2 item remains queued or sender and recipient versions differ, its signature covers a different preimage even though `v` and the wire shape are unchanged. HPKE opening succeeds, but verification returns `identity-signature-invalid`, so a legitimate mailbox item cannot be processed.

@greptile-apps

This P1 is not applicable since there is currently no live implementation of V2, so there are no existing v2 items in the first place.

FSM1:
@coderabbitai review

FSM1:
@coderabbitai full review

FSM1:
@coderabbitai full review

FSM1:
@coderabbitai full review

FSM1:
@coderabbitai full review

FSM1:
@coderabbitai full review

PR #834, opened 2026-07-27
fix: fail closed on short OPFS staging reads and writes and request persistent storage

### LINKED ISSUE #833: client: OPFS staging store ignores short read/write results and never requests persistent storage
Three durability/correctness defects in `packages/client/src/seams/stagingStore.ts`, the OPFS half of the `StagingStore` seam. Surfaced by the #829 staging-budget research on map #813; all three are live today and get worse as the write plane lands on top of this seam.

## 1. `putStagedBytes` discards the short-write result (`stagingStore.ts:91`)

```ts
handle.truncate(0);
handle.write(staged, { at: 0 });   // <- return value dropped
handle.flush();
```

`FileSystemSyncAccessHandle.write()` returns the number of bytes actually written. On a quota-exhausted or otherwise constrained write it can return **fewer bytes than requested without throwing**. Today that truncated block is staged silently, `stagedBytesTotal()` under-counts it, and the drain later uploads a short block whose CID does not match the CID the engine computed for the full bytes.

This is the AGENTS.md rule 8 shape at the staging boundary: the read path hard-rejects a body whose bytes do not match its CID (`crates/engine/src/content/read.rs:224-226`), so the produce side must fail closed rather than stage bytes its own reader will always reject. Note the #821 publish-port design does not help here — its preflight is envelope-level and never sees content blocks.

**Fix:** compare the returned count against `staged.byteLength` and throw on a short write. Delete the partial file before throwing so a retry does not resume from a truncated file.

## 2. `stagedBytes` discards the short-read result (`stagingStore.ts:114`)

```ts
const size = handle.getSize();
const out = new Uint8Array(size);
handle.read(out, { at: 0 });        // <- return value dropped
return out;
```

Same shape on the read side: `read()` returns the bytes actually read, and a short read leaves the tail of `out` as zeroes. The caller cannot distinguish that from legitimately-staged data.

**Fix:** compare the returned count against `size` and throw on a short read.

## 3. `navigator.storage.persist()` is never called anywhere

There is no `persist()` call in `packages/client/src` or `apps/web/src`. Without persistent-storage permission, Safari's ITP evicts unvisited origins' storage after seven days — which deletes **the entire durable op queue and every staged byte**, not just cache. That directly contradicts the offline-parity premise the staging design rests on (`crates/engine/src/sync/staging.rs:4-9`): a user who queues work offline and returns must find their queue intact.

**Fix:** request persistence at engine/worker init, before the first enqueue. `persist()` returning `false` is not fatal — durability is best-effort on some browsers — but the result should be surfaced so the host can warn rather than silently promise durability it does not have.

## Acceptance

- Short write and short read each fail closed with a typed error; the partial file is removed on a short write.
- Persistence is requested once at init and the granted/denied result is observable to the host.
- Unit tests in `packages/client` covering both short-count paths with a fake sync access handle. Note the web unit runner only picks up `*.test.ts` — a `.spec.ts` is silently skipped in CI.

Surfaced by #829. Part of #655.

FSM1:
Closes #833

Three durability defects in the OPFS half of the `StagingStore` seam, all fail-closed fixes at the staging boundary.

## Short write no longer stages a truncated block

`FileSystemSyncAccessHandle.write()` reports a quota-constrained write as a short count, not a throw. `putStagedBytes` dropped that count, so a truncated block was staged silently and the drain would later upload bytes whose CID the engine read path always rejects — the AGENTS.md rule 8 shape at the staging boundary. The returned count is now compared against the copied payload length; a mismatch deletes the partial file (so a retry starts from empty rather than resuming a truncated one) and throws a typed `StagingIoError`. If the cleanup itself fails, the short-write error still wins and carries the cleanup failure as its `cause`.

## Short read no longer returns zero-padded bytes

`read()` reports the same way, and a short read left the tail of the output buffer zeroed — indistinguishable from staged data. The count is now compared against the file size and a mismatch throws `StagingIoError`. The type is module-private and carries no error code: the wasm seam bridge flattens every rejected seam call to a message-only `SeamError`, so a code string would never be observable.

## Persistent storage is requested at engine init

Nothing in the repo called `navigator.storage.persist()`. Without the grant a browser may evict the origin's storage wholesale — the entire durable op queue and every staged byte, not just cache — which breaks the offline-parity premise the staging design rests on.

The request lives in `EngineClient` construction, in the **UI realm**. That placement is load-bearing and was found by verification, not by reading: `StorageManager.persist()` is `[Exposed=Window]`, so a probe of a real Chromium dedicated worker reports `persisted()` present and `persist()` absent — `TypeError: storage.persist is not a function`. An engine-worker-side request, which is where this started, would have thrown and reported denied forever. The grant is origin-wide, so requesting it from the UI covers the worker's IndexedDB and OPFS all the same, and every tab learns the answer, follower tabs included.

The helper is best-effort: it checks `persisted()` first so a browser that prompts does not re-prompt in every tab, and reports `false` rather than throwing when the API is absent or the request fails. The result reaches the host through the new `onStoragePersistence` config callback; a host callback that throws is routed to the existing `onError` seam rather than becoming an unhandled rejection.

## Verification and gates

- Real-browser probe of `StorageManager` in both realms (the finding above), plus the compiled `requestStoragePersistence()` driven in a real Chromium window realm on a secure-context origin: returns a boolean with no throw. Headless Chromium denies the grant on engagement heuristics, so the `true` branch is not reproducible headlessly.
- `pnpm --filter @cipherbox/client test:browser` — 19 passed, including `stagingStore passes its engine conformance kit` against real OPFS in a worker realm.
- `pnpm lint`, `pnpm typecheck`, `pnpm build` — clean.
- `pnpm test` — `packages/client` 73 passed, `apps/api` 155 passed. New unit coverage: short write drops the partial file and closes the handle, cleanup failure still reports the short write, short read rejects, and the persistence helper across missing-API / already-persistent / grant / deny / throw.

No crypto- or privacy-adjacent files in the diff, so `/crypto-privacy-review` does not apply. `/security-review` returned no findings. `/simplify` findings folded in: the transport promise was derived rather than manually settled, the unused seam error code and barrel export were dropped, and duplicated rationale comments were collapsed to one home. CodeRabbit CLI: 1 finding, fixed and covered by a test.

## Summary by CodeRabbit

* **New Features**
  * Added support for requesting persistent browser storage and reporting whether it was granted.
  * Storage persistence errors are safely reported through the client’s error handling.

* **Bug Fixes**
  * Improved staged storage handling to detect incomplete reads and writes.
  * Partial files are removed after failed writes, preventing corrupted staged data.

### DIFF
--- a/packages/client/src/engineClient.test.ts
+++ b/packages/client/src/engineClient.test.ts
@@ -235,4 +235,25 @@ describe('EngineClient leadership + transport swap', () => {
     await expect(inFlight).rejects.toThrow(/retry|closed/);
     await follower.dispose();
   });
+
+  it('reports the leader worker storage-persistence grant to the host', async () => {
+    const locks = new FakeLockManager();
+    const bus = new FakeBus();
+    const seen: boolean[] = [];
+    const leader = new EngineClient({
+      locks,
+      createChannel: () => bus.channel(),
+      spawnWorker: () => {
+        const worker = new FakeEngineWorker();
+        setTimeout(() => worker.ready(true), 0);
+        return worker;
+      },
+      onStoragePersistence: (persisted) => seen.push(persisted),
+    });
+    await tick();
+    await tick();
+
+    expect(seen).toEqual([true]);
+    await leader.dispose();
+  });
 });
--- a/packages/client/src/engineClient.test.ts
+++ b/packages/client/src/engineClient.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from 'vitest';
+import { describe, expect, it, vi } from 'vitest';
 
 import { EngineClient, type EngineClientConfig } from './engineClient.js';
 import { FakeBus, FakeEngineWorker, FakeLockManager } from './testkit.js';
@@ -236,24 +236,21 @@ describe('EngineClient leadership + transport swap', () => {
     await follower.dispose();
   });
 
-  it('reports the leader worker storage-persistence grant to the host', async () => {
-    const locks = new FakeLockManager();
-    const bus = new FakeBus();
-    const seen: boolean[] = [];
-    const leader = new EngineClient({
-      locks,
-      createChannel: () => bus.channel(),
-      spawnWorker: () => {
-        const worker = new FakeEngineWorker();
-        setTimeout(() => worker.ready(true), 0);
-        return worker;
-      },
-      onStoragePersistence: (persisted) => seen.push(persisted),
+  it('reports the origin storage-persistence grant to the host, follower tabs included', async () => {
+    vi.stubGlobal('navigator', {
+      storage: { persisted: () => Promise.resolve(false), persist: () => Promise.resolve(true) },
     });
-    await tick();
+    const { tab } = origin();
+    const seen: boolean[] = [];
+    const leader = tab({ onStoragePersistence: (persisted) => seen.push(persisted) });
+    const follower = tab({ onStoragePersistence: (persisted) => seen.push(persisted) });
     await tick();
 
-    expect(seen).toEqual([true]);
+    expect(leader.currentRole()).toBe('leader');
+    expect(follower.currentRole()).toBe('follower');
+    expect(seen).toEqual([true, true]);
     await leader.dispose();
+    await follower.dispose();
+    vi.unstubAllGlobals();
   });
 });
--- a/packages/client/src/engineClient.test.ts
+++ b/packages/client/src/engineClient.test.ts
@@ -253,4 +253,21 @@ describe('EngineClient leadership + transport swap', () => {
     await follower.dispose();
     vi.unstubAllGlobals();
   });
+
+  it('routes a throwing storage-persistence callback to onError, not an unhandled rejection', async () => {
+    vi.stubGlobal('navigator', { storage: { persist: () => Promise.resolve(true) } });
+    const { tab } = origin();
+    const errors: Error[] = [];
+    const client = tab({
+      onStoragePersistence: () => {
+        throw new Error('host blew up');
+      },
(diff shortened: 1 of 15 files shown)


FSM1:
@coderabbitai review

FSM1:
CodeRabbit review triage: **1 finding, fixed** in `07b54ae01`.

**`putStagedBytes` short-count comment and missing throw coverage** — accepted, and rated higher than Minor on inspection. The finding was in the review body rather than an inline thread, so there is no thread to resolve.

The fix handled a *short count* but not a *throw*. `FileSystemSyncAccessHandle.write` is specified to raise `QuotaExceededError` when storage limits prevent the operation, and that path escaped through the `finally` before the cleanup block ever ran — leaving the partial file staged, since `truncate(0)` has already landed by then. That is the same defect class this PR exists to fix, reached through a different failure mode.

Both signals now converge on a single `failure` value, so the partial file is dropped either way. When the write throws, the original storage error is kept as the `cause`; a cleanup failure only becomes the cause on the short-count path, where there is no original error to preserve. That precedence is pinned by a test rather than left implicit.

The comment claiming short counts happen "without throwing" was corrected — it stated a falsehood about the API contract.

Coverage added: throws from `truncate`, `write`, and `flush`, each asserting the partial file is removed and the handle closed, plus the case where dropping the partial file also fails. The existing short-count tests are untouched — they are the control for the other branch.

Gates re-run on the fix commit:

```
pnpm --filter @cipherbox/client test:browser -> 19 passed (3.6s)
pnpm --filter @cipherbox/client test         -> 10 files, 77 passed (was 73)
pnpm lint                                    -> clean
pnpm typecheck                               -> apps/web, packages/client, apps/api all Done
```

Issue #867, opened 2026-07-28
engine: drain failure policy for dead-letter, blocked, and abandonment


FSM1:
What the drain does when a publish will not succeed. Slice settled by #825 on wayfinder map #813.

## Scope

**Dead-letter classification and reporting.**

- `ReplayReport::dead_letters` reaches the facade. `dead_letters` is filled only from the cold-start undecodable sweep today (`facade.rs:1016`).
- `DeadLetterReason` is carried to the host — none of the four variants reaches it today (#859).
- The valve is indefinite retry on availability, plus dead-letter on a classified-permanent failure or an exhausted attempt budget.
- An `Unconfirmed` op that exhausts its budget dead-letters **without** #819's abandonment retire.

**Abandonment retire** (#819, as amended by #824). Retire fires on abandonment only — dead-letter, cancel, or rebase drop — and must retire an abandoned create's child *name* too. The staged bytes are the retire record: zero drain attempts release immediately, otherwise retire then release. Retire idempotency is a pinned property.

**The `Blocked` class for an over-quota 413** (#841).

- A third class that keeps its place at the head of the queue and its staging reservation. Holding the head is not a freeze: commands still journal, the overlay still renders intent, and reads are untouched.
- Entered on **positive evidence only** — `QUOTA_EXCEEDED` in the 413 body. An absent or unknown code stays permanent. `ApiError::Status` parses only `message` today (`api/client.rs:574-582`) and must learn the `code` discriminator #848 already ships server-side.
- Left when a `GET /account/quota` probe on the drain tick says there is room, giving `pre_flight_quota_check` its first production caller. Never by re-upload: `rawUploadBody` buffers the whole body before the gate runs (`app-setup.ts:37-70`), costing 1 MiB per attempt forever.
- `Command::ManualRefresh` is the retry button, and it already exists.
- Read out of `SnapshotView`, never an event — #824's lossy-event tolerance breaks for a state that *clears*, and a lost "resumed" strands the UI permanently.
- The remedy is behind the blockage: delete, purge and prune are all queued write-plane ops, so retry-after-user-action arrives with no reachable user action.

**A fourth over-budget rejection reason.** On desktop that is an errno, not prose — `blueprint/desktop.md:184-187` collapses both causes into `ENOSPC` when the honest one is `EDQUOT` (noted on #647).

## Acceptance

[...]

Issue #655, opened 2026-07-20
v2 build tracker


FSM1:
The v2 build tracker. Milestone [v2.0.0](https://github.com/FSM1/cipher-box/milestones), label `v2-build`. The blueprint corpus (`blueprint/*.md`, `CONTEXT.md`) is normative; `blueprint/testing.md` law 1 applies to every slice — its suite blocks merges in a named CI job the day it lands.

This tracker is a **sub-issue tree, not a checkbox list**. Every slice lives under one of the nine children below and its state rolls up automatically. Nothing is tracked in this body.

## The tracks

| Track | Scope |
| --- | --- |
| #989 core | The `crates/core` codec, crypto suite, frozen KDF edge catalog, envelope and seal, grant section, IPNS records, content sealing |
| #990 api | The `apps/api` residual surface — identity auth, pin and name registry, hosted upload, mailbox, in-process republisher |
| #991 engine: foundation | Seams and conformance kits, the API client and contract suite, the adoption gate and floor law, resolve and publish, the sync core, grants, rotation, facade cold-start |
| #992 engine: write plane | The frozen content format, op-record rewrap, publishing every op kind, drain failure policy, content upload, vault settings, pin providers, prune |
| #1702 engine: hardening | Engine follow-ups filed after a review gate or a CodeRabbit pass, once #991, #635, and #992 reached the 100 sub-issue cap |
| #993 web | `crates/wasm` bindings and the WASM KAT gate, browser seams, engine worker hosting, tab leadership, the Service Worker byte pipe, the web shell and views |
| #994 desktop | Hardware gates, the Tauri shell, the desktop seam set, the `crates/fuse` operation core, the FUSE-T SMB, libfuse3, and WinFsp adapters, mount lifecycle |
| #995 ci | The main e2e gate and cross-client matrix, the nightly tier, the performance and load harnesses, staging build config |
| #996 deploy | The staging cutover to the v2 stack, release management at the first v2.0.0 RC, the v2 Grafana dashboards |
| #719 retro-audit | Cross-cutting simplify and minor-hardening backlog from the retro sweep of shipped v2 PRs — sits on no slice's critical path |

## Critical path

Core is the critical path. `api` and the desktop hardware gates can start immediately; parallel tracks are fine wherever each issue's own dependencies allow. The write plane was designed on wayfinder map FSM1/cipher-box-next#813 and decomposed by FSM1/cipher-box-next#825.

[...]

Issue #635, opened 2026-07-20
engine: implement rotation primitives, sweep, name wave, and grants


FSM1:
## Scope

The three primitives with no recovery machinery: rotateScope (override seed mint, eager set law — owner cascade over every transitively-reachable descendant scope root fully rotated vs grantee flat self-contained, grant blobs re-wrapped verbatim for the committed set, owner blob, history link, ascent link, structure-signed by the rotator's writer pseudonym); sweep (idempotent epoch-lag work-list, epoch-converged pre-grant convergence, direct-child-scope index self-heal, metadata-only re-seal, idle-cadence Scheduler job, CAS-safe concurrent sweepers); rotateScopeWrite (commitment re-sign, write override seed, background child-first name wave with root re-pointed last, three-channel re-point publish, inventory swap via register-first and batch retire, root linger). The full trigger table including full-depth scope-exit detection, immediate-cut read revoke, write revoke/downgrade, discovered link expiry, manual rotate-now; grant creation (converge, mint scope at epoch 1, write-scope cut, parent index update, publish, mailbox share pointer) and invites (ephemeral-key blobs, bearer write-link flagging, claim conversion).
Does NOT land: scheduled hygiene rotation (designed-for, deferred).

## Blueprint anchors

`blueprint/engine.md`: Rotation primitives; Grants and ledger (grant creation, invites); Pointer planes (re-point publish)

## Gate

Engine simulation: the trigger table and eager-set law mirrored 1:1, sweep idempotence and concurrent sweepers, resumed name waves via the write-plane history link, crash-recovery from published records only, immediate cut, forgery-window bounds, grant-creation convergence requirement.

## Depends on

None — #628 (PR #692), #632 (PR #706) and #633 (PR #735) have all landed.

**Conditional: #902.** If this slice wires the unknown-field-preserving re-encode path (`decode_map_partial` / `encode_map_partial`) into the grant-blob re-wrap or owner/history/ascent re-seal, it becomes the first out-of-crate caller of that codec — and #902 must land first. Today an `UnknownFields` handed across the crate boundary cannot be zeroized by its owner, so the secret bytes are dropped un-wiped and nothing in CI catches it. If the implementation does not touch the partial codec, drop this edge.

## Blocks

- #764 (engine: reclaim orphaned interior write-names after a post-flip write-rotation crash) — its parent primitive is this slice's `rotateScopeWrite` name wave
[...]

PR #897, opened 2026-07-29
feat(engine): add a combined move intent op so a kernel rename is one entry

### LINKED ISSUE #884: engine: add a combined move intent op so a kernel rename is one atomic facade command
## Problem

`blueprint/desktop.md` "Reads, writes, and the never-block law" says a rename becomes **exactly one** facade intent op. It does not: the facade has no combined move-and-rename command, so `crates/fuse`'s `OperationCore::rename` spells one kernel rename as up to three separately-staged ops — `Delete` the destination, `Relink`, `Rename` — with a `?` between each.

The destructive one runs first, and it has to: the engine auto-suffixes name collisions on rebase, so relinking before vacating the destination would land `target (2).txt` instead of a replace.

That leaves a data-loss window. If `Delete` succeeds and `Relink` or `Rename` then fails — durable-queue I/O surfacing as `EngineError::Seam`, or a shutdown mid-sequence — the call returns `Err` and the caller believes nothing happened, but the destination is permanently gone and the source never moved. POSIX guarantees an observer always sees either the old or the new destination.

Secondary: the invalidation pushes all sit after the last `?`, so on a partial failure the kernel keeps a cached positive dentry for a node that has been deleted, for up to the mount's entry TTL.

Found by the `/security-review` and `/crypto-privacy-review` gates on #647.

## Fix

Add a single combined intent op to the facade — shape roughly:

```rust
Command::Move {
    node: NodeId,
    new_parent: NodeId,
    new_name: String,
    replacing: Option<NodeId>,
}
```

staged as one durable op-queue entry, so the whole rename is atomic and the kernel ack means the whole rename is journaled. Then collapse `OperationCore::rename` onto it and delete the three-op sequence.

Needs a rebase rule for the combined op, and a test that a mid-sequence seam failure leaves the destination intact.

## Notes

- The `crates/fuse` side is small once the facade op exists; the work is engine-side.
- Do not attempt the interim reordering — putting `Relink`/`Rename` before the `Delete` trades the loss window for an auto-suffixed wrong name on the success path, which is worse.

Blocks #648 — its ack-at-journal law needs each FUSE operation to be exactly one facade intent op.

Part of #655

FSM1:
Closes #884
Part of #655

## Problem

`blueprint/desktop.md` says a kernel rename becomes exactly one facade intent op. It did not: the facade had no combined move-and-rename command, so `crates/fuse`'s `OperationCore::rename` spelled one kernel rename as up to three separately-staged ops — `Delete` the destination, `Relink`, `Rename` — with a `?` between each. The destructive one ran first and had to (the engine auto-suffixes name collisions on rebase, so relinking before vacating the destination lands `target (2).txt` instead of a replace).

If `Delete` succeeded and a later op failed, the call returned `Err` and the caller believed nothing happened — but the destination was permanently gone and the source never moved. POSIX guarantees an observer always sees either the old or the new destination.

## Fix

`Command::Move { node, new_parent, new_name, replacing }` → `OpKind::Move { from_parent, new_parent, new_name, replacing }`, staged as **one** durable op-queue entry. `OperationCore::rename` collapses onto it and the three-op sequence is gone. The invalidation pushes stay after the single `?`, which is now correct by construction: there is no partial-failure state for a cached positive dentry to survive into.

`replacing` carries a `Replaced { node, sequence }` — the facade snapshots the replaced node's own record sequence, the same anchor `Delete` uses.

### Rebase rule

`rebase_move` runs the relink rule's races first, then the rename rule's collision resolution, over a destination it vacates first:

1. `TargetGone` / `DestinationGone` / `DestinationInsideTarget` dead-letter — one `relocation_guards` shared with `rebase_relink`.
2. A current parent that is neither `from_parent` nor `new_parent` is `MoveRaceLost`.
3. The move vacates the destination node only while that node is **still the one holding the contested name**: same parent, name collates equal, and not advanced past the conditional-delete anchor. A node a concurrent writer renamed or moved away is a bystander — the name is free and the move simply takes it. A move never replaces its own target.
4. Already at the destination under the entered name with nothing to vacate is `AlreadySatisfied`.
5. `resolve_name` runs with both the target and the node being vacated excluded — which is what makes a replace land under the entered name instead of `target (2).txt`. It resolves **before** any mutation, so a `SuffixExhausted` dead-letter leaves the working base as it found it. Its `exclude` parameter became a slice to carry the second exclusion.

### Publish plan

Rename, relink, and move now share one plan, `publish_ref_move`, driven by a `MovePlan`. It keeps the reference-ordering law: dest-add before source-remove, and when source and destination are the same folder the vacated ref and the moved ref land in a **single** record — so a replacing rename in place is one publish and the destination name is never observably unresolvable. The one destructive step re-checks, against the record the gate just handed it, that the ref it drops still holds the name the move is taking.

`compensate_dest_add` takes a `DestAdd` and restores the ref the dest-add vacated — but only in the versioned-safe branch, where the observed sequence proves our own bytes are still the head and the undo is an exact inverse of our own edit. On a `Conflict` it subtracts our add and stops, since a winner has already built on the listing we published.

## Tests

- `crates/fuse`: a durable-queue outage never destroys the destination a rename did not replace, at journal budgets 0 and 1 — verified red against the pre-change three-op sequence (`budget 1: the destination name vanished`). Plus: `rename(a, a)` journals nothing, and a replaced junk-holding folder keeps its destination entry through a mid-sweep outage.
- `crates/engine` rebase: replace lands under the entered name; a concurrent edit that saves the replaced node forces the auto-suffix; a bystander renamed **or** moved out of the way survives (both verified red without the guard); a self-replacing move keeps its target; rename in place does not churn the link counter; the race loser and both dead-letter reasons leave the working base untouched.
- `crates/engine` overlay: one `Move` relinks, renames and replaces in one step, and keeps a target named as its own replaced node.
- `crates/engine` write-plane (drain, end to end): a replacing rename lands the vacated and moved refs in one record and a second device resolves it; a cross-folder replacing move still publishes dest before source; a compensated move restores the ref its dest-add replaced (verified red without the restore) and does **not** resurrect it over a concurrent winner.

`cargo fmt --all`, `cargo clippy --workspace --all-targets` (zero warnings), and `cargo test --workspace` (31 suites) are green, plus `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown` since the `Command` enum changed.

## Review gates

All three ran on `git diff main...HEAD`; findings are folded in as the three follow-up commits.

- `/simplify` — `publish_move` duplicated ~85% of `publish_relink`, and the rebase and overlay arms restated the relink+rename bodies. Now one `MovePlan`/`publish_ref_move`, one `relocation_guards`, one overlay `relocate`, one `relocation_anchors` in the facade, and a shared `delete_descendants` in the fuse crate. Also caught a real regression: the collapsed `rename` was staging an op for a no-op `rename(a, a)`.
- `/security-review` — found the vacate was anchored on the replaced node's record sequence alone. A name lives in the parent's child ref, so renaming a node never advances that node's own sequence; a concurrently-renamed destination read as unchanged and the bystander was unlinked. Fixed in the rebase and re-checked in the drain.
- `/crypto-privacy-review` — found the self-replace hole (`replacing == node` erased the target, dead-lettering every later op on it) and the compensation's additive undo replaying over a `Conflict` winner. Both fixed. No new crypto surface: the diff adds no `kdf::*`/`seal_*`/entropy/signer calls, every record still reaches the plane through the unchanged `publish_node` chain with one fresh injected nonce per seal, and no filename, node id, or sequence reaches an error, event, or log.

## Notes for review

- Expect a rebase on `main` before merge: #867 lands in `drain.rs`, `facade.rs`, and `api/client.rs`. This PR adds a publish path rather than restructuring `Halt`, the `pass`/`run` loop, or the attempt-budget surface.
- No failure-policy behaviour here. `publish_ref_move` fails with the same bare `Halt` every other plan uses, so #867's classification applies to it unchanged with no new seam.
- Cross-scope move stays out: `OpKind::Move` carries no `cross_scope`/`exits_granted_source`, exactly as the facade stages `Relink` today. That is #635's. A cross-scope destination cannot produce a mis-keyed record — `ensure_folder` requires the chain to reach the scope root, so it halts.
- `Command::Rename` and `Command::Relink` stay — the web client drives them as separate user actions — but they no longer have publish plans of their own.
- Residual, deliberately not closed here: when the destination is a folder holding hidden platform junk, those descendants still get their own deletes before the move, and the engine may then decline to vacate the destination (conditional delete, or a bystander). The POSIX-visible entry always survives — only junk the mount hides can be lost. Covered by `replacing_a_junk_holding_folder_keeps_the_destination_entry_when_the_queue_fails`.
- One markdown line in `blueprint/engine.md` changed emphasis style on an untouched line: the file was previously outside lint-staged's reach, and both prettier and markdownlint demand it the moment the file is edited.

## Summary by CodeRabbit

* **New Features**
  * Added unified move operations that combine relinking, renaming, and optional destination replacement in one step.
  * Publishing updates ensure replacement moves appear atomically and preserve moved node identity.

* **Bug Fixes**
  * Improved rename/move compensation so vacated destination entries are restored correctly without resurrecting replaced refs after concurrent winners.
  * Added safer handling for self-renames and durable-queue interruption scenarios to keep directory entries consistent.

* **Documentation**
  * Updated engine blueprint documentation to reflect the unified move model.

* **Tests**
  * Expanded write-plane and failure-mode tests covering atomic publishing, replacement semantics, and compensation correctness.

### DIFF
--- a/blueprint/desktop.md
+++ b/blueprint/desktop.md
@@ -176,7 +176,9 @@ there is no `block_on` freeze of the whole mount behind one slow call.
   plaintext `cb-write-*` exposure class is gone (judgment; the zero-overwrite
   cleanup papers over what this closes structurally).
 - **release** (and create/mkdir/rename/unlink/rmdir) — becomes exactly one
-  facade intent op (`updateContent`, `create`, `rename`, `relink`, `delete`).
+  facade intent op (`updateContent`, `create`, `move`, `delete`). A kernel
+  rename is one `move`, which carries the relink, the rename, and the
+  destination it replaces, so a replace is never observable half-done.
   The kernel is acked when the op is **journaled durably in the StagingStore**
   — the v1 INV-1 no-false-ack discipline, now uniform across every mutation
   instead of two. Everything after the ack (seal, upload, publish, CAS rebase)
--- a/blueprint/engine.md
+++ b/blueprint/engine.md
@@ -218,11 +218,11 @@ poll timer, desktop from FUSE-op TTL checks — the core is identical.
   exactly two things: trust violations and an empty-cache cold start. Manual
   refresh resolves with nocache semantics everywhere.
 - **Ops**: every mutation is an intent op — `create`, `delete`, `rename`,
-  `relink`, `updateContent` — carrying its base sequence and its authored
+  `relink`, `move`, `updateContent` — carrying its base sequence and its authored
   time, journaled FIFO in the durable op queue (all mutations, both platforms)
   as a versioned, owner-tagged record whose intent body seals HPKE-to-self in
   **auth mode**, so authoring one requires the owner's enc secret rather than
-  the public tag stamped beside it. That authenticates a record's *author*, not
+  the public tag stamped beside it. That authenticates a record's _author_, not
   its freshness or its position: a store co-tenant can still copy, delete, or
   reorder whole records, which queue-integrity work covers separately. The queue
   is per device, not per account, and is shared with whatever build wrote it: a
@@ -232,7 +232,10 @@ poll timer, desktop from FUSE-op TTL checks — the core is identical.
   dead-lettered and dropped. An intra-scope
   `relink` is a pure relink; a cross-scope `relink` re-seals the moved subtree
   at the destination scope's epoch, and one that leaves a granted source scope
-  is a scope-exit rotation trigger for the source (#26 D1/D7). Replay is FIFO
+  is a scope-exit rotation trigger for the source (#26 D1/D7). A `move` is a
+  relink and a rename in one entry, optionally vacating the node already at the
+  destination name — one POSIX rename is exactly one `move`, so the whole
+  operation is journaled or none of it is. Replay is FIFO
   in performed order through the standard rebase, and rebases only onto
   gate-passing state (#33 D5–D7).
 - **Withheld-update escalation**: shared scopes only — a name pinned past a
--- a/crates/engine/src/facade.rs
+++ b/crates/engine/src/facade.rs
@@ -321,8 +321,9 @@ pub enum Command {
         /// Destination parent.
         new_parent: NodeId,
     },
-    /// Relink and rename a node in one intent op, replacing whatever sits at
-    /// the destination name ([`OpKind::Move`](crate::OpKind::Move)).
+    /// Relink and rename a node in one intent op, conditionally replacing the
+    /// node at the destination name — a concurrent edit to it wins and the move
+    /// auto-suffixes instead ([`OpKind::Move`](crate::OpKind::Move)).
     Move {
         /// Node being moved.
         node: NodeId,
--- a/crates/engine/src/facade.rs
+++ b/crates/engine/src/facade.rs
@@ -322,8 +322,7 @@ pub enum Command {
         new_parent: NodeId,
     },
     /// Relink and rename a node in one intent op, replacing whatever sits at
-    /// the destination name. One kernel rename is exactly one of these
-    /// (blueprint/desktop.md "Reads, writes, and the never-block law").
+    /// the destination name ([`OpKind::Move`](crate::OpKind::Move)).
     Move {
         /// Node being moved.
         node: NodeId,
@@ -1484,10 +1483,7 @@ impl<T: SeamTypes> Engine<T> {
             }
             Command::Relink { node, new_parent } => {
                 let rendered = self.render().await?;
-                let from_parent = rendered
-                    .parent_of(node)
-                    .unwrap_or(self.snapshot.borrow().root);
-                let base_sequence = rendered.record_sequence(node).unwrap_or(1);
+                let (from_parent, base_sequence) = self.relocation_anchors(&rendered, node);
                 // trailing bools: cross_scope=false, exits_granted_source=false — intra-scope pure relink
                 let op = Op::relink(
                     node,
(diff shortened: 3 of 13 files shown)


FSM1:
## Review resolution

One finding, from the CodeRabbit review body (no review threads were opened on this PR).

**`crates/engine/src/facade.rs:324` — "Docs overstate the replace as unconditional." — VALID, fixed in 75be125ab.**

Verified against the implementation before rewording:

- `rebase_move` (`crates/engine/src/sync/rebase.rs`) only vacates the destination when the replaced node is still under `new_parent`, still collates equal to `new_name`, and has not advanced past the recorded conditional-delete sequence. A node a concurrent writer renamed or moved away is a bystander; a node a concurrent edit advanced wins, `vacating` stays `None`, and `resolve_name` therefore suffixes the move beside it.
- The drain agrees: the `OpKind::Move` arm sets `MovePlan.vacated` only for the node the rebase actually vacated, so a winner keeps its child ref.

New wording:

```rust
/// Relink and rename a node in one intent op, conditionally replacing the
/// node at the destination name — a concurrent edit to it wins and the move
/// auto-suffixes instead ([`OpKind::Move`](crate::OpKind::Move)).
```

The full conditional-vacate rule stays stated once at its home on `rebase_move`; the variant doc keeps a one-line summary plus the existing cross-reference.

Doc-only change. Gates re-run green: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (all suites pass, 0 failures), `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown`.

Issue #868, opened 2026-07-28
engine: write file content end to end


FSM1:
Content bytes flow: sliced by the client, sealed and staged by the engine, uploaded at drain, published as a version. Gives `ApiClient::upload` its first production caller and the contract suite its first live write coverage. Slice settled by #825 on wayfinder map #813.

## Scope

**Facade write handles** (#815). The client slices the `File` and feeds chunks; the engine seals and stages each, owning framing policy; the op is journaled once at commit. Peak wasm heap is O(chunk). This replaces `Command::Create { content: Option<PlaintextContent> }` (`facade.rs:238-247`).

**Budget admission** (#828, #829). Reserve at `beginWrite` against an exact sealed total (`size + 40n + root_len`) in an in-memory ledger of live handles, so concurrent handles cannot over-admit. `pushChunk` enforces the declared shape and never re-checks the budget. The over-budget check quotes **`budget - staged`**, never `budget`, in distinct messages for the hard limit, a transient backlog, and a device-constrained budget — they need different user actions.

**Per-block staging** (#828). A version stages as N+1 independent blocks, each keyed by its own content address; the op carries the root's, which is also the version's `contentCid`. Leaf keys are derivable by decoding the staged root. Blocks are **CID-verified on read from staging** — the read path already hard-rejects a mismatch, so an unverified upload turns bit-rot into a permanently unreadable published version.

**The per-version content key** (#818). Staged sealed, HPKE-to-self under the owner `enc-subkey`, riding the op. `enc-subkey` comes from the loginSecret, so it is epoch-independent and available exactly when a drain can run. Epoch rides the AAD as a value, not a key input. Needs a distinct struct tag and the `contentCid` in the sealed payload. A key-open failure dead-letters the op and **releases** its blocks.

**Drain upload.** File order, root last, removing each block on a confirmed `UploadResult`, so present blocks are always a **suffix** — a hole is loss, not progress, and fails closed. Upload then publish; do not pre-register CIDs, since `NameRegistration` has no CID-only form and would mint a null-head name-inventory row on every new-file write.

**The remaining two encode-side guards** (#823), both structural: one `NodeKind` feeds both the sealed body variant and `ChildRef.kind`; `SealedContent` gains `size` so the `Version` is built from one value.

**`content_cids` threaded into held records** (#797) so `eol_renew_pass` re-pins content on sub-EOL renewal.

[...]

Issue #812, opened 2026-07-23
engine/core: encode-side fail-closed guards for the child read-path rejects


FSM1:
## Context

PR #811 (issue #802) added four decode/verify-side hard rejects on the child read path whose matching produce paths do not exist yet. Per the AGENTS.md encode/decode fail-closed symmetry rule, each producer must land with a release-active `Err` guard (never `debug_assert!`) plus a test that fires in a release build — otherwise a release build can publish bytes its own decoder rejects.

## Debts

When the content-bearing write / child-publish paths land, each of these decode-side rejects needs its encode-side twin:

1. **Grant section on a child envelope** — child adoption rejects any envelope carrying a `grantSection` key (`has_grant_section`, `crates/core/src/seal/envelope.rs`). Child envelope sealing must refuse to attach one.
2. **Kind transplant** — `read_content` rejects when the sealed body kind disagrees with the parent's `ChildRef.kind` (`crates/engine/src/facade.rs`). Child-ref authoring must enforce the match.
3. **Version size vs manifest size** — `content::open_content` rejects `manifest.size != version.size` (`crates/engine/src/content/mod.rs`). Today `Version::new` and DAG `assemble` take `size` independently; seal/publish must tie them.
4. **Child-ref `ipnsName` well-formedness** — the read path parses the bytes fail-closed as a canonical base36 IPNS name (`crates/engine/src/facade.rs`); child-ref authoring must emit exactly that form.

## Dependency

Depends on the engine content-bearing write / upload slice #868, which declares `Closes #812`; blocks none. Related precedent: #794. Sibling AGENTS.md rule 8 debt, disjoint work: #893.

Found by the /crypto-privacy-review gate on #811.

FSM1:
Bidirectional edge: depends on the content-bearing write path surfacing via #807.

FSM1:
Unblocked and fully specified by #823 (map #813). All four guards land in the **engine's record-authoring layer**; `crates/core` gains no new surface.

| # | Guard | Form | Mechanism |
| --- | --- | --- | --- |
| 1 | Grant section on a child envelope | **checked** | Child-authoring rejects release-active via core's existing `has_grant_section` (`crates/core/src/seal/envelope.rs:91`) |
| 2 | Kind transplant | **structural** | One `NodeKind` feeds both the child's `ReadBody` variant and the parent's `ChildRef.kind` |
[...]

PR #901, opened 2026-07-29
fix(core): size the encode_map_partial buffer up front

### LINKED ISSUE #893: core: convert encode_map_partial to the two-pass sized buffer for encode-realloc hygiene
## Context

#681 closed the codec-wide encode-realloc hygiene gap by making `codec::encode` two-pass: `encoded_len` sizes the buffer up front so the single allocation is exact and the write never reallocates. `crates/core/src/codec/encode.rs` states the rule in its own rustdoc:

> A grow-from-empty `Vec` frees each intermediate backing store un-zeroized, stranding secret bytes in freed heap; one right-sized buffer keeps them to the terminal owner's plaintext, which that owner zeroizes.

## Gap

`encode_map_partial` in `crates/core/src/codec/fields.rs` was not converted. It still builds its output with `Vec::new()` and grows it:

```rust
pub fn encode_map_partial(known: &Map, unknown: &UnknownFields) -> Result<Vec<u8>, CodecError> {
    let mut out = Vec::new();
```

It then writes the known field values through `write_value(&mut out, value, 1)` on that growing buffer. Those values are re-built known fields of a decoded structure — on the read-body path they carry inline content keys, and grant/owner/history seed material as those re-encode paths land. Every reallocation during the append frees the previous backing store un-zeroized, which is exactly the property #681 set out to eliminate.

The bytes on the wire are unchanged; this is purely allocation hygiene, and it is the same class of finding as #681, not a new one.

## Suggested direction

Mirror `encode`'s two-pass shape: size the merged map up front (`known` field lengths via the existing `encoded_len` machinery, plus the preserved `unknown` raw byte lengths and their key heads) and `Vec::with_capacity` once, so no reallocation occurs mid-write. Alternatively route the known-field writes through `encode` and splice, if that lands cleaner than a second length pass.

## Exposure today

`encode_map_partial` is `pub` from `crates/core/src/codec/mod.rs` but is currently only exercised by tests, KAT generation, and the KAT manifest suite — no production caller feeds it secret-bearing fields yet. So this is a **latent** gap to close before the unknown-field-preserving re-encode path is wired into the seal/grant surfaces, not a live leak.

## Dependencies

Depends on nothing — ready to pick up. #681 (the codec-wide two-pass conversion, closed) is the precedent this mirrors.

Blocks nothing hard, but it must land **before** any slice wires the unknown-field-preserving re-encode path into the seal/grant surfaces, since that is the point `encode_map_partial` acquires its first secret-bearing production caller. Sibling AGENTS.md rule 8 debt: #812.

## Provenance

Spotted during the comment-discipline sweep (PR forthcoming) while reconciling `encode.rs`'s stated two-pass rationale against its siblings. Deliberately not fixed there — that PR is comment-only, no logic changes.

Part of #655

FSM1:
## What

`encode_map_partial` was the one encoder left building its output from an empty `Vec`. Every reallocation during the append frees the previous backing store un-zeroized, stranding partially-serialized known-field values in freed heap — the property #681 closed for `codec::encode` and that `crates/core/src/codec/encode.rs` documents in its own rustdoc.

This mirrors `encode`'s two-pass shape, and goes one step further: **every reject now fires before the output buffer exists.**

- `merge` walks the two canonically-ordered sides once, producing the emit sequence and rejecting a key present on both sides. It holds borrows only, so no secret bytes are copied into it, and it is itself sized exactly.
- `merged_len` sizes the merged map exactly, in emit order: the map head, each key's text head + bytes, and either `count_value` of a known value at the emit loop's depth or a preserved unknown value's raw canonical byte length.
- `encode_map_partial` then does a single `Vec::with_capacity` and writes.

Both rejects — `unknown-field-collision` and `depth-exceeded` — therefore run before a byte is allocated for the output. Previously the collision reject fired mid-write and dropped a buffer already holding written known-field bytes un-zeroized, which is the same leak class this issue is about. Counting in emit order also restores `depth-exceeded`'s reported offset, which a split known-then-unknown sum under-reported.

`count_value`, `head_len`, and `text_len` in `encode.rs` widen from private to `pub(super)` (`super` = `codec`, which is a private module) — no public API surface added and no wire-format change.

### Why the sizing pass rather than splicing through `encode`

The issue offered "route the known-field writes through `encode` and splice" as an alternative. That is strictly worse for the property being fixed: it would mint one intermediate `Vec` per known value, each holding secret bytes and each dropped un-zeroized. The second length pass keeps the secret bytes to exactly one buffer.

## Wire bytes are unchanged

This is purely allocation hygiene. **No KAT vector was regenerated** and the KAT generator was not run — the `crates/core/kat` tree hash is identical to `main` (`b6b0901a7dad3f4e8198ac02eb35c932fcf00dc7`), and the diff touches three files, none of them vectors. The KAT manifest suite passes unchanged in debug and release, and the `Core KATs (native + WASM)` gate is green.

## Tests

- `merged_write_never_reallocates` (unit): an 80-entry merged map — 40 known, 40 preserved unknown — spanning both key-head size classes (short keys, and 32-byte keys whose text head is two bytes) and a >23-entry map head, with 32-byte secret-shaped values. Asserts `capacity == len` on the buffer the public `encode_map_partial` returns, an allocator-independent `merged_len` sizing oracle against `out.len()`, and byte-equality against the direct `encode`. Verified **red** against the pre-fix allocation: `capacity 5120` vs `len 4562`.
- `merged_encode_rejects_past_max_depth` (unit): the merged path's `depth-exceeded` reject, release-active per AGENTS.md rule 8. The partial encoder had no coverage of it on either side of this change.
- Codec property suite (b) `unknown_field_round_trip` gains the same `capacity == len` assertion, fuzz-covering `merged_len` against the emit loop across arbitrary maps. Also verified red against the pre-fix allocation.

## Gates

`cargo fmt --all --check`, `cargo clippy --all-targets`, `cargo test --workspace`, `cargo test --release -p cipherbox-core`, and `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown` all green.

Review gates run on `git diff main...HEAD`: `/simplify`, `/security-review`, `/crypto-privacy-review`. Findings folded in: the collision-reject leak, the `depth-exceeded` offset regression, the missing depth-reject test, and the allocator-dependent sizing witness. Both reviewers independently confirmed `merged_len` is exact head-for-head against the emit loop, that the `usize` sum cannot overflow, that `pub(super)` adds no public surface, and that the change introduces no timing signal.

Out of scope: #812 (encode-side fail-closed guards for the child read-path rejects) is disjoint sibling AGENTS.md rule 8 debt, owned by #868. The pre-existing gap that `UnknownFields` and `Map` give out-of-crate callers no way to wipe secret-bearing values is filed as #902.

Closes #893
Part of #655

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved unknown fields during partial map decoding and re-encoding.
  * Ensured canonical, byte-stable output for partially decoded maps.
  * Detected excessive nesting before encoding begins.
  * Prevented unnecessary buffer reallocations during map encoding.
* **Tests**
  * Added coverage for unknown-field round trips, exact buffer sizing, canonical output, and nested-data validation.

### DIFF
--- a/crates/core/src/codec/encode.rs
+++ b/crates/core/src/codec/encode.rs
@@ -54,7 +54,7 @@ pub fn encoded_len(value: &Value) -> Result<usize, CodecError> {
     Ok(len)
 }
 
-fn count_value(len: &mut usize, value: &Value, depth: usize) -> Result<(), CodecError> {
+pub(super) fn count_value(len: &mut usize, value: &Value, depth: usize) -> Result<(), CodecError> {
     check_depth(depth, *len)?;
     match value {
         Value::Unsigned(n) | Value::Negative(n) => *len += head_len(*n),
@@ -89,12 +89,12 @@ fn check_depth(depth: usize, offset: usize) -> Result<(), CodecError> {
 }
 
 /// The byte length of a text item: its head plus its UTF-8 bytes.
-fn text_len(t: &str) -> usize {
+pub(super) fn text_len(t: &str) -> usize {
     head_len(t.len() as u64) + t.len()
 }
 
 /// The byte length of a shortest-form head for `arg` (mirrors [`write_head`]).
-fn head_len(arg: u64) -> usize {
+pub(super) fn head_len(arg: u64) -> usize {
     match arg {
         0..=23 => 1,
         24..=0xff => 2,
--- a/crates/core/src/codec/fields.rs
+++ b/crates/core/src/codec/fields.rs
@@ -217,19 +217,34 @@ mod tests {
 
     /// AGENTS.md rule 8: the merged encoder's `depth-exceeded` reject is
     /// release-active, and now fires from the sizing pass — before a byte is
-    /// allocated for the output.
+    /// allocated for the output. The deep value sits behind two preserved
+    /// unknowns, so its reported offset only matches the equivalent full-map
+    /// encode's if the sizing pass counted in emit order.
     #[test]
     fn merged_encode_rejects_past_max_depth() {
+        fn depth_offset(err: &CodecError) -> usize {
+            match err {
+                CodecError::Malformed(Malformed::DepthExceeded { offset }) => *offset,
+                other => panic!("expected depth-exceeded, got {}", other.check()),
+            }
+        }
+
         let original = encode(&Value::Map(sample_map())).unwrap();
         let (mut known, unknown) =
-            decode_map_partial(&original, known_key_set(&["v", "id"])).unwrap();
+            decode_map_partial(&original, known_key_set(&["v", "zz-later"])).unwrap();
         let mut deep = Value::Null;
         for _ in 0..crate::codec::MAX_DEPTH {
             deep = Value::Array(vec![deep]);
         }
-        known.insert("v", deep);
-        let err = encode_map_partial(&known, &unknown).unwrap_err();
-        assert_eq!(err.check(), "depth-exceeded");
+        known.insert("zz-later", deep.clone());
+        let mut full = sample_map();
+        full.insert("zz-later", deep);
+
+        assert_eq!(
+            depth_offset(&encode_map_partial(&known, &unknown).unwrap_err()),
+            depth_offset(&encode(&Value::Map(full)).unwrap_err()),
+            "merged reject offset diverged from the full-map encode"
+        );
     }
 
     #[test]
--- a/crates/core/src/codec/fields.rs
+++ b/crates/core/src/codec/fields.rs
@@ -12,7 +12,7 @@ use super::decode::Decoder;
 use super::encode::{
     MAJOR_MAP, count_value, head_len, text_len, write_head, write_text, write_value,
 };
-use super::value::{Map, canonical_key_cmp};
+use super::value::{Map, Value, canonical_key_cmp};
 use crate::error::{CodecError, Malformed};
 
 /// The depth a top-level map's values sit at: the map itself is the enclosing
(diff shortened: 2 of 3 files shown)


FSM1:
## CodeRabbit review — resolved

Rebased onto `main` first (3 commits behind: #898, #899, #897), clean replay, then addressed the one finding. Posting here rather than on a thread: the review was body-only, with zero review threads to resolve.

### `crates/core/src/codec/fields.rs:218-233` — "Assert the restored depth-error offset" — **Accepted, fixed**

Valid. Nothing pinned the offset: `merged_encode_rejects_past_max_depth` asserted only `err.check() == "depth-exceeded"`, and `encode::tests::encode_rejects_past_max_depth` likewise checks only the name on both passes. The offset that `check_depth` documents as "the emitted-byte position the bound fired at" was unasserted everywhere.

Took the suggested comparison against the equivalent full-map encode, which pins the value without hardcoding it. One change beyond the literal suggestion: the deep value moved from `"v"` (first in canonical order) to `"zz-later"` (last), so it now emits *after* two preserved unknowns. On the old shape both the correct emit-order count and the buggy known-then-unknown count produce the same offset, so the assertion would have been vacuous.

Mutation-checked that it bites: restoring the pre-fix "sum known side, then unknown side" order in `merged_len` fails the test with `left: 140, right: 158`, while `merged_write_never_reallocates` still passes — the total length is unchanged under that reordering, so this assertion is the only thing that catches it.

### Wire bytes

Unchanged. `crates/core/kat` tree hash is `b6b0901a7dad3f4e8198ac02eb35c932fcf00dc7`, identical to `main`; no vector added or regenerated. The commit is test-only.

### Gates

All green on the rebased head: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, `cargo test --release -p cipherbox-core`, `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown`.

Issue #871, opened 2026-07-28
engine: wire the pin-provider layer across all three modes


FSM1:
Wires `PinMode` across all three modes and fixes the advisory-quota bypass. `PinMode`, `ByoIpfsConfig` and `pre_flight_quota_check` are re-exported from `lib.rs:51-54` with zero callers anywhere. Slice settled by #825 on wayfinder map #813.

## Scope

**`PinMode` is a client-side placement policy; the provider layer is concrete dispatch over `Http`; the hosted leg is authoritative in both directions** (#822).

- The API hard-refuses hosted ingress for a BYO account under a row lock (`apps/api/src/content/content.service.ts:246-248`), so `users.byo` is two-state while `PinMode` is three-state and **Dual has no server representation**. Therefore `byo=true` is exactly `External`; `Hosted` and `Dual` both run `byo=false`. That is correct accounting, not a workaround, and it needs no API change.
- **Dual runs both legs, both retrying inside the op, but only hosted can fail it.** Under strict-FIFO stop-at-first-failure, a both-must-succeed rule would let an offline home Kubo stall every later mutation in the vault.
- `ByoIpfsConfig` is read from the settings record.
- The vaulted mode is the source of truth, reconciled against the `advisory` field the command-time quota fetch already returns.

**#705's fix.** `pre_flight_quota_check` short-circuits `Ok(())` on a single server-provided `advisory` boolean. Gate on the **local byte path** — check for `Hosted | Dual`, skip for `External` — and demote `quota.advisory` to a display hint. The advisory flag must track the byte path, not the account.

**The pre-flight is in, at command time only.** It can never be authoritative (#826's four miss-modes), so it goes where the seal-and-stage wait already is, sized in **sealed** bytes.

**Dual's external-leg partial outcome** — hosted succeeded, the member's provider did not — is reported per-op rather than swallowed.

## Acceptance

- An `External` account's write skips the hosted byte path and the hosted pre-flight.
- A `Dual` write with an unreachable provider publishes and reports the partial outcome.
- A `Hosted` write over quota fails fast at command time rather than at drain.

## Gate

`Engine Tests`, `Contract Suite`.

## Dependencies

Blocked by the content write slice (#868) and the settings record slice (#870).

[...]

Issue #653, opened 2026-07-20
deploy: execute the staging cutover to the v2 stack


FSM1:
## Scope

The one scheduled, announced, destructive redeploy: stop the v1 stack, remove the postgres and ipfs volumes, deploy the v2 stack fresh — no data migration, no dual-running. Finishes the deploy-staging.yml edit (gateway forward_auth wiring: Caddy in front of Kubo's gateway port, forward_auth to the API token-verification endpoint, volume-wipe cutover support); docker-compose.staging.yml per the service table (redis and tee-worker die); dashboards rewritten in their own slice (#656); secret retirements at cutover (STAGING_TEE_WORKER_SECRET, PHALA_*, STAGING_REDIS_PASSWORD, SENDGRID_*); the deploy-landing env-name mismatch fixed. Cutover trigger: v2 contract suite + smoke green on a VPS dry-run.
Accepted consequence, stated: existing v1 desktop installs auto-update into v2 and their local vaults become orphans — the v2 first-run treats them as new devices.

## Blueprint anchors

`blueprint/deploy.md`: Staging pipeline (stack table, gateway shape, cutover); Secrets and environments

## Gate

The v2 stack serving on staging after the destructive redeploy; the WIP-deploy workflow_dispatch path exercised at least once before the cutover.

## Depends on

- #906 (api: pin uploaded blocks under their caller-computed content address) — until this lands, the hosted accelerator stores blocks under Kubo's own UnixFS address instead of the engine-computed content address, so published records resolve to bytes the accelerator does not hold and `publish_record` fails closed on every metadata write. A staging stack cannot serve a read path in that state.

FSM1:
## Cloudflare origin-pull state is provisioned and waits for this cutover

PR #1526 is merged as `b906203db`, so `docker/Caddyfile` on `main` imports the `(origin_tls)`
snippet. That snippet sets `mode require_and_verify` and pins the trust pool to one file. **The
deploy of that Caddyfile is the step that arms it, and this issue owns that deploy.**

The Cloudflare side and the VPS side are already done, on 2026-08-31. The state is inert until the
deploy: Cloudflare presents the client certificate, the running origin does not request one, and all
4 staging hosts answer 200 today.

### What is in place

| Item | State |
| --- | --- |
| Zone | `cipherbox.cc`, id `05d71f72b3699f643d33c191e64c56c5` — the same zone the `staging` variable `CLOUDFLARE_ZONE_ID` names |
[...]

PR #912, opened 2026-07-31
fix(api): pin uploaded blocks under their caller-computed content address

### LINKED ISSUE #906: api: pin uploaded blocks under their caller-computed content address
The hosted ingress stores a block under **Kubo's own UnixFS address**, not under the content address the engine computed and registers. Every published record therefore points at a CID the accelerator does not hold.

## Evidence

`KuboPinStore` uploads through `/api/v0/add` with no CID parameters (`apps/api/src/registry/pin-store.ts`), so Kubo returns a CIDv0 `dag-pb` sha2-256 hash. Against the local stack:

```text
POST /api/v0/add?only-hash=true  (15 bytes)
-> {"Hash":"Qme9fM8s5TKq2RsJj2mRZfKn7JWf9cViQimAfriisRa3Hd"}
```

The engine addresses every block as CIDv1 base32 with a BLAKE3-256 multihash — `raw` for content leaves, `dag-cbor` for DAG roots and record head blocks (`crates/core/src/content/cid.rs`, `crates/engine/src/content/dag.rs`). The two can never be equal.

## Consequences

1. **Records are unreadable through the accelerator.** A published IPNS record's value is `/ipfs/<headCid>` where `headCid` is the engine's `dag-cbor` address; the bytes sit in Kubo under a `Qm...` address instead, so the gateway GET the read path makes returns nothing.
2. **`publish_record` fails closed forever.** It compares `uploaded.cid` against the head block's own address and returns `RecordPublishError::HeadCidMismatch` on disagreement (`crates/engine/src/net/record_publish.rs`), which the drain classes as availability and retries indefinitely. No metadata write can confirm against real infrastructure.
3. **Registered content CIDs name nothing.** `contentCids` carries engine-computed addresses, so the pin rows and the pinned bytes are keyed differently and refcounting cannot line up.

This predates the content write slice — the metadata publish path has the same dependency on the head block's address — but it was unreachable until #868 gave `ApiClient::upload` a production caller. The engine tests pass because the fake block plane indexes a block under its own content address; the live contract suite passes because it asserts the API's lifecycle contract and never compares the returned CID to a caller-computed one.

## What resolving looks like

The caller already knows the address it needs. Options, in rough order of preference:

- **The client declares the CID** on `POST /content/upload`, and the API pins with `block/put` under the declared multicodec (`cid-codec` + `mhtype=blake3&mhlen=32`), failing closed if Kubo's returned address disagrees. Removes the API's pre-commit `PinStore::hash` guess entirely — the declared CID is what takes the advisory lock — and keeps CID computation out of TypeScript.
- The API derives the codec from the block, which it cannot do without interpreting the bytes.

Decide the surface, update `blueprint/api.md`, and extend the contract-suite leg to assert the returned address **equals** the declared one — the assertion this issue exists because nothing can make today.

## Dependencies

Surfaced by #868 (the content write slice), which gave the upload path its first production caller.

Blocks:

- #871 (engine: wire the pin-provider layer across all three modes) — it dispatches the same byte path across hosted, external, and dual modes, so its hosted leg is broken until this is fixed.
- #653 (deploy: execute the staging cutover to the v2 stack) — a staging stack cannot serve a read path while every published record points at a CID the accelerator does not hold.

Part of #655

FSM1:
Closes #906. Unblocks #871 and #653.

## The defect

`KuboPinStore` uploaded through `/api/v0/add`, so Kubo returned a CIDv0 `dag-pb` sha2-256 hash. The engine addresses every block as CIDv1 base32 over a BLAKE3-256 multihash — `raw` for sealed content leaves, `dag-cbor` for DAG roots and record heads. The two can never be equal, so every published record pointed at a CID the accelerator does not hold and `publish_record` failed closed forever on `HeadCidMismatch`.

## The surface

The issue's preferred option, confirmed workable against the local Kubo v0.42.0 before it was built:

- `POST /content/upload` carries the caller's own content address in an **`X-Content-Cid` header**. A header rather than a query parameter: content addresses correlate across accounts, and edge proxies log request URLs — so they stay out of the URL plane exactly as the registry's `contentCids` stay inside JSON bodies. It also keeps URL interpolation out of the engine client entirely.
- The pin store writes the block with `block/put?cid-codec=<from the declared CID>&mhtype=blake3&mhlen=32&pin=false`, compares Kubo's returned `Key` to the declared string, and only then pins it. A disagreement is a permanent **400**, not a retriable 503 — the declaration is wrong and retrying the same body can never succeed.
- `PinStore.hash` is gone. The API computes no content address of its own; the declared CID is what takes the per-CID advisory lock, keys the pin row, and is echoed back. Kubo's re-derivation is what binds bytes to CID — the declared string is only a routing hint until it matches.
- The API reads the multicodec out of the frozen CIDv1 framing by prefix (`bafkr4i` -> `raw`, `bafyr4i` -> `dag-cbor`) rather than decoding or recomputing anything, so no codec or crypto lands in TypeScript. Anything outside those two frozen shapes is refused before any lock or quota work. The mapping cannot be a trust hole: the codec is embedded in the address Kubo returns, so a wrong one simply fails the comparison.
- **Every path that will not pin the block removes it.** `block/put` persists before the address can be checked, and an unpinned block is not reclaimed — the daemon runs without `--enable-gc`. So a refused upload cleans up after itself; Kubo refuses to remove a *pinned* block, so the cleanup can never release another account's content.
- **One request, one block.** `MAX_UPLOAD_BYTES` now defaults to the 2 MiB `block/put` ceiling, so an over-size body is a permanent 413 (`UPLOAD_TOO_LARGE`, which the drain dead-letters) rather than a retry-forever 503.

### Direct pins, not recursive

`block/put --pin` produces a **recursive** pin, and Kubo's repo GC then walks it. A `dag-cbor`-codec block whose bytes are not decodable CBOR aborts GC for the whole node — reproduced locally: `could not retrieve links for bafyr4i...: unexpected content after end of cbor object; garbage collection aborted`. Any authenticated account could have parked one there. Blocks are therefore written unpinned and pinned **direct** via `pin/add?recursive=false`, which is also the correct shape for this plane: every block is uploaded and registered individually, and Kubo itself rejects an undecodable `dag-cbor` block at `pin/add`.

## What the contract assertion now proves

`upload_pins_under_the_caller_computed_content_address` runs in the merge-blocking `contract-suite` job against a real API and a real Kubo. The load-bearing proof is a **gateway round-trip**, not the echoed CID — the API answers with the address it was handed, so only fetching `/ipfs/<declared>?format=raw` back and byte-comparing shows the block really landed there:

- a sealed leaf under `raw` and a real assembled DAG root under `dag-cbor` (genuine det-CBOR bytes) are each served back byte-identical at the address the caller computed;
- bytes that do not hash to the declared CID are refused with a permanent 400 **and** the account's quota usage is unchanged afterwards, so the compensation is proven, not assumed;
- a non-canonical CID, and a canonically-framed CID under a codec the ingress does not accept, never leave the client.

The job gains `CONTRACT_GATEWAY_URL`; the leg fails loudly if it is missing rather than skipping silently (verified by unsetting it).

New unit coverage pins the pin store's actual Kubo effects: `block/put` -> `pin/add` on success, `block/put` -> `block/rm` on a mismatch, `block/rm` after a failed pin, and a put that returns no address reported as a store fault rather than a caller mismatch.

## Review gates

`/simplify`, `/security-review` and `/crypto-privacy-review` all ran against `git diff main...HEAD`. Folded back in:

- **A refused upload grew the datastore for free** (HIGH, both reviews). The block was written before the compare and never removed, while the quota charge was refunded — an authenticated account could add ~120 MiB/min of permanent, unmetered storage. Fixed by removing the block on every non-pinning path.
- **`MAX_UPLOAD_BYTES` was 50x the block ceiling**, turning a permanent size failure into a retryable 503 that re-ran the row insert and wrote another orphan on each attempt.
- **The declared CID sat in the URL**, exporting per-account content-address and timing edges to the edge log plane. Moved to a header.
- A `block/put` response with no `Key` (Kubo can trail an error object under a 200) read as a disagreeing address and blamed the caller with a permanent 400; it is a store fault now.
- The client-side guard accepted any single-byte multicodec while the ingress accepts two — narrowed so a wider codec breaks at the client instead of as a production 400.
- A failed compensation logs at `error`: it strands a charged row that makes later uploads of those bytes short-circuit as already-pinned.
- `RecordPublishError::HeadCidMismatch`'s doc claimed the returned CID proves what bytes the store holds. With an echoed CID that is no longer true, and the crypto review flagged it as a stale in-code claim; the doc now says what the check actually means. The bytes/CID binding is enforced at the ingress and, client-side, by `verify_cid` on read — unchanged.
- Comment trimming: the GC rationale and the frozen-shape derivation are stated once at their home in `blueprint/api.md` rather than restated on three callers.

## Verification

- `cargo test --workspace` green; `cargo clippy --workspace --all-targets -D warnings` clean; `cargo fmt --all`.
- `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown` green.
- API unit suite 177 passed, integration suite 132 passed against real Postgres; `tsc --noEmit`, eslint, prettier and markdownlint clean.
- Contract suite 15/15 against the local stack, including the gateway round-trip; `apps/api/openapi.json` regenerated for the `openapi-freshness` gate.

## Deliberately out of scope

- **#915** — the engine authors head blocks and DAG roots up to 4 MiB (`MAX_RESOLVED_RECORD_BYTES`) while IPFS caps a block at 2 MiB. Not currently reachable (sealed leaves are exactly 1 MiB), and the ceiling is a frozen wire-format limit owned by the content slice, so this PR fixes only the API-side half. Filed with a dependency edge back here.
- The drain classifies a 400 as `Halt::Unclassified`, so a mis-declared CID retries rather than dead-letters. Only reachable from a client bug; publish-failure classification belongs to the content-write slice.
- The API does not validate that `dag-cbor` bytes are decodable — Kubo's `pin/add` already refuses them, and doing it in the API would mean a CBOR codec in TypeScript.

## Collision note

#907, for #868, touches `crates/engine/src/content/`, `crates/core/src/seal/`, and `crates/contract/tests/contract.rs`. Engine changes here are confined to `api/client.rs`, `api/error.rs` and one call site in `net/record_publish.rs`; everything else is expressed API-side, including the block-ceiling fix. The contract-suite additions are appended and do not touch the areas #907 edits, but expect a rebase there.

## Summary by CodeRabbit

* **New Features**
  * Content uploads now require a caller-provided content address.
  * Raw and DAG-CBOR addresses are validated, and successful uploads confirm storage under the declared address.
  * Added end-to-end integrity checks, including gateway retrieval verification.

* **Bug Fixes**
  * Invalid addresses and content mismatches return clear client errors.
  * Failed uploads clean up incomplete data without charging quota.

* **Changes**
  * Uploads are limited to 2 MiB (one block).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -738,6 +738,9 @@ jobs:
       CONTRACT_API_URL: http://localhost:3000
       CONTRACT_API_PROD_URL: http://localhost:3100
       CONTRACT_TEST_LOGIN_SECRET: contract-suite-secret
+      # The stack's gateway, so the suite can fetch back a block the ingress
+      # pinned and prove it landed under the caller-computed address (#906).
+      CONTRACT_GATEWAY_URL: http://localhost:8080
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -615,6 +615,7 @@ dependencies = [
 name = "cipherbox-contract"
 version = "0.0.0"
 dependencies = [
+ "cipherbox-core",
  "cipherbox-engine",
  "getrandom 0.2.17",
  "reqwest 0.13.4",
--- a/apps/api/openapi.json
+++ b/apps/api/openapi.json
@@ -18,9 +18,7 @@
           }
         },
         "summary": "Liveness check",
-        "tags": [
-          "Ops"
-        ]
+        "tags": ["Ops"]
       }
     },
     "/metrics": {
@@ -40,9 +38,7 @@
           }
         },
         "summary": "Prometheus metrics in text exposition format",
-        "tags": [
-          "Ops"
-        ]
+        "tags": ["Ops"]
       }
     },
     "/auth/challenge": {
@@ -72,9 +68,7 @@
           }
         },
         "summary": "Issue a single-use login challenge bound to an identity publicKey",
-        "tags": [
-          "Auth"
-        ]
+        "tags": ["Auth"]
       }
     },
     "/auth/login": {
@@ -104,9 +98,7 @@
           }
         },
         "summary": "Challenge-signature login against the secp256k1 identity key; creates the account implicitly at first login",
-        "tags": [
-          "Auth"
-        ]
+        "tags": ["Auth"]
       }
     },
     "/auth/siwe/challenge": {
@@ -126,9 +118,7 @@
           }
         },
         "summary": "Issue a single-use SIWE nonce",
-        "tags": [
-          "Auth"
-        ]
+        "tags": ["Auth"]
       }
     },
     "/auth/siwe/login": {
@@ -158,9 +148,7 @@
           }
         },
         "summary": "SIWE wallet login (secondary method; the wallet must already be linked)",
-        "tags": [
-          "Auth"
-        ]
+        "tags": ["Auth"]
       }
     },
     "/auth/siwe/link": {
@@ -195,9 +183,7 @@
           }
         ],
         "summary": "Link a SIWE wallet to the authenticated account",
-        "tags": [
-          "Auth"
-        ]
+        "tags": ["Auth"]
       }
     },
     "/auth/refresh": {
@@ -227,9 +213,7 @@
           }
         },
         "summary": "Rotate the refresh token (one-time-use; reuse revokes the whole family)",
-        "tags": [
-          "Auth"
-        ]
+        "tags": ["Auth"]
       }
     },
     "/auth/logout": {
@@ -254,9 +238,7 @@
           }
         ],
         "summary": "Revoke every refresh token for the account (hard delete)",
-        "tags": [
-          "Auth"
-        ]
+        "tags": ["Auth"]
       }
     },
     "/auth/test-login": {
@@ -286,9 +268,7 @@
           }
         },
         "summary": "Staging-gated deterministic login for e2e (hard-blocked in production)",
-        "tags": [
-          "Auth"
-        ]
+        "tags": ["Auth"]
       }
     },
     "/mailbox/messages": {
@@ -344,9 +324,7 @@
           }
         ],
         "summary": "Post an HPKE-sealed blob to a recipient identity publicKey; unknown recipients are rejected (rate-limited existence oracle)",
-        "tags": [
-          "Mailbox"
-        ]
+        "tags": ["Mailbox"]
       },
       "get": {
         "operationId": "MailboxController_poll",
@@ -375,9 +353,7 @@
           }
         ],
         "summary": "Poll the authenticated mailbox; returns sealed blobs with no sender metadata in the clear",
-        "tags": [
-          "Mailbox"
-        ]
+        "tags": ["Mailbox"]
       }
     },
     "/mailbox/messages/{id}": {
@@ -417,9 +393,7 @@
           }
         ],
         "summary": "Ack a message: hard delete by id, scoped to the caller mailbox",
-        "tags": [
-          "Mailbox"
-        ]
+        "tags": ["Mailbox"]
       }
     },
     "/registry/register": {
@@ -470,9 +444,7 @@
           }
         ],
         "summary": "Batch register [{ipnsName, headCid?, contentCids[]}] under the caller account; register-first, idempotent upserts",
-        "tags": [
-          "Registry"
-        ]
+        "tags": ["Registry"]
       }
     },
     "/registry/retire": {
@@ -524,9 +496,7 @@
           }
         ],
         "summary": "Batch retire [ipnsName | cid] for the caller account; union liveness, refcounted physical unpin at global zero",
-        "tags": [
-          "Registry"
-        ]
+        "tags": ["Registry"]
       }
     },
     "/account/quota": {
@@ -554,9 +524,7 @@
           }
         ],
         "summary": "The caller per-account quota: usedBytes is the gated sum the upload gate enforces, pinnedBytes the all-rows sum, limitBytes the override or env default, advisory the BYO flag",
-        "tags": [
-          "Account"
-        ]
+        "tags": ["Account"]
       }
     },
     "/account/byo": {
@@ -597,9 +565,7 @@
           }
         ],
         "summary": "Toggle the caller BYO-IPFS flag; advisory pin rows follow it",
-        "tags": [
-          "Account"
-        ]
+        "tags": ["Account"]
       }
     },
     "/account": {
@@ -630,9 +596,7 @@
           }
         ],
         "summary": "Immediate hard-delete of the caller account: retire inventory/pin rows (refcounted unpin), purge mailbox, delete auth rows",
-        "tags": [
-          "Account"
-        ]
+        "tags": ["Account"]
       }
     },
     "/content/upload": {
@@ -640,9 +604,9 @@
         "operationId": "ContentController_upload",
         "parameters": [
           {
-            "name": "cid",
+            "name": "x-content-cid",
             "required": true,
-            "in": "query",
+            "in": "header",
             "description": "The caller-computed content address of the body: a CIDv1 base32 BLAKE3-256 CID whose codec is `raw` (sealed content leaf) or `dag-cbor` (DAG root or record head). The pin store rejects the upload if the bytes do not address to it.",
             "schema": {
               "type": "string"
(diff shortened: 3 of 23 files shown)


FSM1:
Rebased onto `main` (now at #907) and resolved the review.

**CodeRabbit nitpick — accepted.** The only actionable item was the body-only nitpick on `content.http.integration.test.ts` (no review thread to resolve): every HTTP upload declared a `raw`-codec CID, so the controller's `dag-cbor` acceptance was never exercised at the HTTP layer. Added a success-path test for it, and taught the fake pin store to read the codec back off the declared address the way Kubo does, rather than assuming `raw`.

**Semantic drift the rebase exposed.** `git` reported this branch MERGEABLE — there was no textual conflict — but it did not compile against current `main`. Two independent breaks, both invisible to the merge:

1. **Arity.** #907 added `sync/drain.rs::upload_block`, which called `api.upload(block)` with one argument. This branch's diff never touched `drain.rs`, so nothing updated it. Threaded the block's staging key (its content address) through `upload_block` and encoded it at the API boundary, keeping the drain in its native byte key type. Also fixed a second orphan call site in `crates/contract/tests/contract.rs`.
2. **Type change.** #907 changed `content::assemble` from `&[SealedChunk]` to `&[Vec<u8>]` of leaf CIDs. The contract suite's new upload leg still passed the old shape.

**A test fake that was wrong in this branch's favour.** The write-plane HTTP fake asserted the declared address against `put()`, which addresses every block under the `dag-cbor` codec — so it rejected a correctly declared `raw` leaf. Replaced with a put-and-compare that accepts an address under either frozen content-plane codec, mirroring the real ingress. Production code was correct; only the fake was wrong.

Also corrected a stale doc on `ApiError::MalformedContentCid`, which said the address is interpolated into the request URL. It rides a header — that is the whole point of `X-Content-Cid`.

Verified on the rebased head: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo check --workspace --all-targets`, `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown --all-targets` all clean; `cargo test -p cipherbox-engine -p cipherbox-core` 19/19 binaries green; API unit 177/177 and integration 133/133 green. The Contract Suite needs a live API instance and was not run locally — CI is authoritative for it.

FSM1:
The latest review's single item is an outside-diff-range comment on `publish_update_content` (`crates/engine/src/sync/drain.rs`): include retained versions' leaf CIDs in the content registration, flagged Major / Data Integrity.

**Verdict: not applying it. No code change, and no live durability defect.**

The premise is right and worth stating plainly: pins are **direct**, not recursive — `pin/add?arg=...&recursive=false` (`apps/api/src/registry/pin-store.ts`), deliberately, per `blueprint/api.md` "Blocks are pinned direct, not recursive". So pinning a `dag-cbor` root genuinely does not retain the leaves it links. The conclusion drawn from it is what does not hold.

Registration is not what keeps a retained version durable:

- `register` **never pins and never unpins**. It upserts `name_inventory` and does an idempotent `INSERT` into `pinned_cids` — no `DELETE`, no tombstone, no retirement marker (`apps/api/src/registry/services/registry.service.ts`).
- `pinned_cids` is keyed `(accountId, cid)` with **no name column**, so there is no per-name registered set that a later registration could shrink by omission.
- Leaves are pinned once, at upload (`content.service.ts`), and are released only by an explicit `retire` target list. There is no background EOL/GC sweep that reclaims unregistered CIDs, and Kubo runs without `--enable-gc` anyway.

So a retained version's leaves stay pinned whether or not a later republish re-names them. Adding them would be a no-op against the database — `register` filters already-known CIDs out before insert.

It would not be a no-op on the wire, and this is the part that matters: `contentCids` is capped at **1000 per entry**, enforced fail-closed with a 400 (`@ArrayMaxSize(MAX_CONTENT_CIDS)`, `apps/api/src/registry/dto/registry.dto.ts`). Issue #920 already tracks that the engine's caller is unbounded and that a refused registration does not dead-letter — it loops forever at the strict-FIFO queue head, blocking every op behind it. Widening the set from O(leaves) to O(versions x leaves) would make that reachable for ordinary files after a few edits, converting a large-file edge case into a routine one. The suggested "quick win" would worsen a tracked fail-closed defect. I have added the reasoning to #920 so it is not re-proposed there.

One genuine adjacent gap did surface while tracing this, and it is **already tracked**: `registered_by` (`drain.rs`) retires only `op.content_root_cid()` on abandonment, never the leaves the same publish registered — despite its own doc claiming it mirrors `PublishRequest::registration()`, which sends root plus every leaf. Since retire is explicit-list-driven and nothing GCs, those leaves stay pinned and keep counting against quota. That is #916, already open and out of scope here.

No commit pushed; head remains d486b35dcf28d5da2963036eb9a0d3aef117f84e.

Issue #917, opened 2026-07-31
engine: nothing resolves a folder record below the scope root on a device that did not author it


FSM1:
## Gap

#895 set out to assert the two-device round trip on a create two levels deep. The
record plane holds, and #895 pins it. The facade cannot express the assertion:
**a device that did not author a subtree renders it empty below the scope root,
indefinitely.**

What is wired today:

- `spawn_resolve_tick_loop` (`crates/engine/src/facade.rs`) resolves the **vault
  root only**. `refresh_base_from_outcome` -> `project_root`
  (`crates/engine/src/sync/project.rs:19`) lifts **direct children only** — its own
  unit test is named `project_root_lifts_direct_children_only`.
- The only other writer into the base below the root is the drain's self-adopt
  (`project_folder`, `crates/engine/src/sync/drain.rs:1636`) — state this device
  published itself.
- `Command::SetFocus` falls through the `command` dispatch to
  `EngineError::Unimplemented`.
- `ChildAdopter` (`crates/engine/src/net/child.rs`) — the child-record gate — has
  exactly one caller, `Engine::read_content`, which is file-only (`NotAFile` for a
  folder) and requires the node to already be in the rendered view.
- `focus_set` / `FocusWindow` / `FocusTarget::Folder`
  (`crates/engine/src/sync/tick.rs`) are pure planning with no caller.

`blueprint/engine.md` "Sync core" specifies the tick as: *"refresh the vault
pointer, the open folder, and its full ancestor chain to root ... Any other cached
folder refreshes on access past the staleness threshold"*. Neither half is wired.

## Evidence

Driving #895's assertion through the facade — `engine_b.snapshot(photos).children`
on a cold second device — yields **0** children, while the parent's published
record carries exactly one. #895 shipped the record-plane half as
`a_create_below_the_scope_root_is_adoptable_by_a_second_device`, which adopts that
same record through `ChildAdopter` on device B's own floors and cache and reads the
depth-2 child out of it. This issue is the facade half.

## Blast radius

- `apps/web/src/engine/snapshotStore.ts:145` already calls
  `client.facade.setFocus(node)` on navigation; it rejects `Unimplemented` and the
  store commits an error state.
- Any second device — web navigation or FUSE `readdir` — lists nothing below the
  root for a tree it did not author.
- Related but distinct: #855 (size/mtime for files a device never downloaded) is
  the same read-plane deficit one field further in.

## Fix

[...]

Issue #916, opened 2026-07-31
engine: retire every uploaded content block on abandonment, not only the version root


FSM1:
## Problem

Every content block a version uploads becomes its own **charged** pin row, and only the version **root** is ever retired. The leaves are never retired on any path, so an abandoned upload permanently spends account quota.

`POST /content/upload` inserts one authoritative row per block — `apps/api/src/content/content.service.ts` inserts `{ accountId, cid, size, advisory: false }`, and `sumHostedBytes` (the quota gate) counts exactly those rows. Registration at publish time only inserts `size: '0'` placeholders (`apps/api/src/registry/services/registry.service.ts`), so the accountable row is created by the upload, not by the registration.

`Drain::upload_version` (`crates/engine/src/sync/drain.rs`) uploads every leaf and then the root, one `api.upload` each. `registered_by` returns only the write name and `op.content_root_cid()`, so `abandon` retires at most the root.

Reachable case: a 40-leaf version whose leaf 39 or whose root upload answers `413 UPLOAD_TOO_LARGE` classifies as `Halt::Permanent(PayloadRefused)`, the op is abandoned, and 39 leaf pin rows stay charged against the account with no reachable record. `release_staged_blocks` simultaneously drops the staged copies, so the client no longer holds the CID list either — the rows are unrecoverable from the client side.

`registered_by`'s comment ("retiring the root is what drops the reference count the accelerator's own GC walks") does not match the API's accounting: `PinnedCid` rows are independent per-CID rows and the server walks no manifest.

A second reachable path never even reaches `abandon`: the `Halt::Attempt` arm of `apply_valve` dequeues and releases staged blocks without calling `abandon` at all, so a version whose publish registered and then exhausted its attempt budget retires nothing.

## Scope

- Enumerate the version's leaf CIDs at abandonment (the staged root block still has them at that point — `release_staged_blocks` already reads it) and add them to the retire batch.
- Decide the batch bound: a version at the flat-DAG ceiling has thousands of leaves, and `POST /registry/retire` takes a bare JSON array with no documented cap. Either chunk the batch client-side or give the endpoint an explicit limit.
- Retire on the `Halt::Attempt` arm too, or route it through `abandon`.
- Correct the `registered_by` comment about root-retirement dropping a reference count.

`retire` is already idempotent for a never-registered target (`registry.service.test.ts` "is idempotent: retiring an already-gone target succeeds as a no-op"), so retiring leaves that never landed is safe.

## Tests

[...]

Issue #805, opened 2026-07-23
web: harvest the v1 vault browser read path over the snapshot adapter


FSM1:
## Goal
Bring back the v1 vault browser read/list/navigate/breadcrumbs, sourced from the snapshot adapter.

## Harvest / create
Harvest from branch `v1`:
- `components/layout/{AppShell,AppHeader,AppFooter,AppSidebar,NavItem,UserMenu}.tsx`
- `components/file-browser/{FileBrowser,FileList,FileListItem,Breadcrumbs,ParentDirRow,EmptyState}.tsx`
- `routes/FilesPage.tsx`
- `styles/{layout,file-browser,breadcrumbs,responsive}.css`
- `utils/{format,fileTypes}.ts`

## Rewire
- `useFolderNavigation` reads `snapshotStore`, keys on `nodeId`, `setFocus` on nav.
- Progressive hydration: name/kind instant, size/mtime later.
- `AppSidebar` stubs `/shared` `/bin` `/settings` as disabled with a coming-soon-#643 affordance; `/files` stays live.
- Projection is direct-children-only per #789 — breadcrumbs use the ancestor chain in `SnapshotView`; no code may assume a materialized full tree.

## Engine seams
`snapshot`, `setFocus`, `subscribe`.

## Tests
Render-from-snapshot; nav updates focus; breadcrumb chain; progressive hydration.

## Gate
web unit.

## Depends on
#803, #900, #917.

## Blocks

#809 — its login-and-CRUD e2e smoke gate drives the vault-browser read path this slice lands, so there is nothing for it to walk until this merges.

Part of #642

FSM1:
Depends on #900 — the snapshot adapter landed in #803 seeds its first pull with a hardcoded all-zero root id because the wasm `snapshot` seam takes no "the root" option. Breadcrumbs and navigation here build on that root handling, so #900 should land before or alongside this slice.

PR #922, opened 2026-07-31
test(engine): assert a deep create is adoptable by a second device

### LINKED ISSUE #895: engine: assert the two-device round trip on a create below the scope root
## Gap

`a_create_below_the_scope_root_publishes_and_projects` (`crates/engine/tests/write_plane.rs:968`) — the test that landed with #887 — is single-device. It asserts against the record store (`published_names`) and against the writing engine's own rendered snapshot, but never cold-starts a second device to resolve the depth-2 create off the published record plane.

#887's scope asked for exactly that: "extend `crates/engine/tests/write_plane.rs` with a create two levels deep, asserting the two-device round trip that is standing from #865 onward". Every other op kind in that file carries the round trip; the deep create is the one that does not, so nothing proves a non-root parent's published record is resolvable by a device that did not author it.

## Fix

Test-only, roughly 20 lines. Add a cold second device to the existing test (or a sibling test): `world.device(b"alice-second-device")`, `serve_http`, `engine_on`, `start`, then resolve the depth-2 create through `snapshot`/`view` and assert the child name and id match device A's. Follow the pattern at `write_plane.rs:597-601` in `a_second_device_of_the_same_account_resolves_the_write`.

No engine change is expected — if the second device cannot resolve it, that is a real finding and the issue grows.

## Gate

`Engine Tests`.

## References

- #887 — the slice that landed the deep-create path and declared this assertion in scope (closed by PR #892).
- #890 — closed as a duplicate; this was the one genuine residual identified there.

## Dependencies

Depends on nothing — the engine path it asserts landed with #887 (PR #892), so this is pickable now.

Blocks nothing. It is test-only in `crates/engine/tests/write_plane.rs`, which #867 and #884 also touch — land it either side of those, or expect a trivial rebase.

Part of #655

FSM1:
## What

Every other op kind in `crates/engine/tests/write_plane.rs` carries a two-device
round trip. The create below the scope root did not, so nothing proved that a
**non-root parent's** published record is resolvable by a device that did not
author it — the scope root's record cannot stand in for it, because a child write
stops at the immediate parent.

`a_create_below_the_scope_root_is_adoptable_by_a_second_device` closes that: device
A creates `photos/` then `photos/2026/`, a cold second device boots on its own
floor store, snapshot cache and HTTP, resolves the depth-1 parent off the root
record through its own cold start, and then adopts `photos`' **own** record through
the child gate — `ChildAdopter` + `resolve`, the same pipeline `Engine::read_content`
uses — and reads `2026` out of the gate-passing body, matching device A's node id.

Test-only. No production code changed.

## The finding: the assertion cannot be made through the facade

#895 asked for the round trip through `snapshot`/`view` on device B. Written that
way it **fails**: `engine_b.snapshot(photos).children` is empty.

That is not a defect in the deep-create publish path — the record plane is correct,
which is exactly what this test now pins. It is a missing read-plane capability:

- `spawn_resolve_tick_loop` resolves the **vault root only**; `refresh_base_from_outcome`
  to `project_root` lifts **direct children only** (its unit test is literally named
  `project_root_lifts_direct_children_only`).
- The only other writer into the base below the root is the drain's own self-adopt.
- `Command::SetFocus` falls through the `command` dispatch to `EngineError::Unimplemented`.
- `ChildAdopter` has exactly one caller, `read_content`, which is file-only.
- `focus_set` / `FocusWindow` / `FocusTarget::Folder` are pure planning with no caller.

So any device that did not author a subtree lists nothing below the scope root.
`apps/web/src/engine/snapshotStore.ts` already calls `facade.setFocus(node)` on
navigation and takes the `Unimplemented` rejection into its error state.

Filed as #917 with the reciprocal edge recorded on #805 whose "Engine seams" list
already names `setFocus`. #917 owns the facade half; this PR ships the record-plane
half.

## Verification

- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo check --workspace --all-targets` — clean
- `cargo test -p cipherbox-engine` — 38 passed in `write_plane`, 0 failed; the new
  test is named in the output. Gate: `Engine Tests` runs `cargo test -p cipherbox-engine`.

The test fails when the invariant it pins is broken — confirmed by deliberate breaks:

| Break | Result |
| --- | --- |
| Drop device A's second drain tick, so the depth-2 create never publishes | fails: `device B reads the depth-2 create, left: 0, right: 1` |
| Point the adopter at the wrong `expected_node` | fails: `TrustViolation(GateRejection { stage: Unseal, reason: Trust(Trust(SealOpenFailed)) })` |
| Bind the adopter to a foreign `scope_id` | fails: same fail-closed `Unseal` rejection |

The last two confirm the adoption gate is genuinely exercised rather than bypassed
with shared in-process state: only the record store, the block plane and the clock
are shared with device A; device B's floors, cache and HTTP are cold and its own.

## Review gates

- `/simplify` — tightened the doc comment to a single cross-reference. Skipped the one
  other finding: the `GatewayConfig` literal duplicates the one inlined in the shared
  `engine_on` helper, and de-duplicating it means editing a shared fixture in a file
  #916 is concurrently adding tests to.
- `/security-review` — no findings. Confirmed every gate stage the assertion claims is
  actually run: record signature verify under the name, real head-block fetch with CID
  verification, envelope id and scope binding, per-name sequence floor against device B's
  cold floor store, AAD-bound unseal.
- `/crypto-privacy-review` — no findings. `ChildAdopter` takes the seed by value and is
  its terminal owner, so the zeroize-at-the-terminal-owner rule holds; no encode/decode
  path is touched, so AGENTS.md rule 8 does not apply; no `ResolveOutcome` variant can
  render key or seed bytes through the panic message.

Closes #895
Part of #655

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when deeply nested folders are created and synchronized across multiple devices.
  - Ensured child folders are correctly adopted and retain their identity when discovered by another device.

- **Tests**
  - Added coverage for cross-device synchronization of nested folder structures, including parent resolution and child visibility.

### DIFF
--- a/crates/engine/tests/write_plane.rs
+++ b/crates/engine/tests/write_plane.rs
@@ -23,6 +23,7 @@ use zeroize::Zeroizing;
 use cipherbox_engine::content::{DAG_ROOT_CODEC, GatewaySource, SealedChunk, decode_root};
 use cipherbox_engine::facade::PendingClass;
 use cipherbox_engine::net::author::{EnvelopeAuthoring, author_child_envelope};
+use cipherbox_engine::net::{ChildAdopter, ResolveOutcome, resolve};
 use cipherbox_engine::seams::{
     BoxedTask, HttpRequest, HttpResponse, OpId, RecordTransport, SeamError, SeamResult,
     StagingStore, UnixMillis,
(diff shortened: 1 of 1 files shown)


FSM1:
@coderabbitai review

FSM1:
@coderabbitai full review

Issue #940, opened 2026-08-01
engine: a corrupt upload mark count is trusted without bound


FSM1:
## Problem

`Drain::upload_mark` (`crates/engine/src/sync/drain.rs`) is fail-closed against a mark naming a *different* version — a root-CID mismatch reads as zero — but it does not bound the count it parses:

```rust
Ok(<[u8; 4]>::try_from(count).map_or(0, |c| u32::from_be_bytes(c) as usize))
```

A mark of `(thisRoot, u32::MAX)` therefore makes every absent leaf skip the hole guard, and the drain publishes a version whose manifest names blocks nothing holds. The staging store is host-local storage, so a torn or bit-rotted mark file that still parses is the reachable source — not an attacker, but not merely theoretical either.

This is pre-existing and predates #924. It is filed separately rather than folded into #924 because the fix is a genuine design choice, not a mechanical bound.

## The choice to make

A count above the version's own leaf count is proof of corruption, but the two repairs point in opposite directions:

- **Treat it as zero progress.** Fail-closed for integrity: no version publishes over blocks nothing holds. But every leaf an earlier pass already uploaded *and released* is now absent and uncovered, so the hole guard raises `CONTENT_LOST`, the valve abandons the op, and a write whose bytes are on the network is destroyed — precisely the outcome #924 exists to prevent.
- **Clamp to the leaf count.** Preserves the write, but that is the maximally trusting reading of bytes already known to be corrupt, and it can still publish an unassemblable version.

A third option — treat a corrupt mark as unclassified and retry, or re-derive progress by asking the pin provider which of the version's CIDs it already holds — costs a round trip but is the only reading that neither destroys the write nor publishes a manifest that will not resolve. That is the option worth pricing before implementing.

## Scope

- Decide the reading, implement it in `upload_mark`, and cover it with a test that plants a corrupt mark directly in the staging store and asserts the chosen behavior.
- Whatever is chosen, a corrupt mark must not silently publish a version whose links do not resolve.

## Provenance

Found by the `/crypto-privacy-review` gate on the #924 fix, which makes the mark more load-bearing: the mark is now the sole record of which released leaves are recoverable, so the weight placed on its contents went up.

## Depends on

- #924 — it establishes the mark as a high-water mark and is the reason this gap is worth closing now. Land after it. **Satisfied:** #924 is closed, so this edge no longer blocks.

Part of #655

Issue #869, opened 2026-07-28
engine: cancel an upload and collect orphaned staged blocks


FSM1:
The lifecycle of staged blocks: the user can abandon an upload, and blocks nobody references are collected. Slice settled by #825 on wayfinder map #813.

## Scope

**Cancel** (#824). `Command::CancelUpload { op_id }`, content-only.

- Content-only because for a metadata op a compensating mutation is already equivalent — rename it back — while for an upload it is not: the compensating delete still pushes the whole file through the network and never returns the staging budget.
- **Guaranteed until publish entry, refused after** with `TooLateToCancel`. Cancel can therefore never mutate published state, and #821's `Unconfirmed` lands on the refused side by construction. Rejected: converting a late cancel into a compensating `Delete`, which silently substitutes an irreversible published mutation.
- Requires a **block-boundary yield** in the drain's upload loop. Without it the guarantee collapses to "cancel only works before the op starts".
- Staged bytes are **released**, under the rule that also covers #818's carve-out: *preserved when the engine gave up on an op, released when the user did*.
- Cascade keys on whether the node survives: a cancelled **create** takes every later queued op on that node; a cancelled `UpdateContent` takes nothing, since versions are independent full writes.
- The host sees `OpPhase::UploadCancelled` on the existing progress event.

**Orphan GC** (#828). Expands staged roots into leaf sets and collects blocks nothing references.

- Freezes the whole pass on a root it cannot expand — self-clearing via the drain's permanent classification.
- Excludes live handles **by CID**: the write handle records a leaf CID *before* staging its bytes.
- Runs at **cold start and after each drain pass**.
- Staged blocks are removed in whole sets or in file order, never individually out of order.

**#853's trap.** A dead-lettered op stops being a GC root, so the cold-start path's own op-record removal makes the *preserved* staged bytes collectible. The four-way staged-byte lifetime must be reconciled in one place: preserved on dead-letter, released on #818's proven-unopenable blob, released on cancel, released on an unexpandable staged root.

## Acceptance

- A cancel mid-upload releases every block of that version and returns the reservation.
- A cancel after publish entry is refused.
- GC leaves a live handle's blocks untouched while its bytes are still being staged.
- A dead-lettered op's blocks survive a cold start and a GC pass.

## Gate

`Engine Tests`.

## Dependencies

Blocked by the content write slice (#868), and by #896 for the `OpPhase::UploadCancelled` variant this slice specifies.

[...]

Issue #819, opened 2026-07-27
wayfinder: what is the upload/register/publish ordering, and the rollback on partial failure?


FSM1:
Part of #813

## Question

What is the exact ordering of upload, register, and publish — and what is the rollback when a step fails partway?

## Why it is a fork

Blocks must be uploaded and pinned before a record is published, or the read path resolves a valid record pointing at unfetchable CIDs. But **register-first** (`CONTEXT.md:71`, `blueprint/engine.md:464-470` — "register-first ordering is built into the publish pipeline, not left to callers") says the name and CID batch registers before first publish. Meanwhile the upload endpoint creates its own pin row. The CIDs are known locally before any bytes move, because `compute_cid` is a local computation (`crates/core/src/content/cid.rs:70`) — so more orderings are available than are obviously correct.

## Candidates

1. `upload(leaves) -> upload(root) -> upload(envelope) -> register(name, contentCids) -> publish`
2. `register(name, contentCids) -> upload -> publish` — true register-first, viable because CIDs are computable before upload
3. Upload-only, letting the upload endpoint's own pin row satisfy inventory — skips name registration and violates register-first, but should be rejected explicitly

## The rollback half

If publish fails after upload, blocks are pinned and quota-charged with nothing referencing them. Decide which applies:

- GC via `retire` (`net/retire.rs:24`, live today)
- the "registered-never-published orphan, alert-GC'd" residual (`CONTEXT.md:71`)
- retry from the still-queued op, since the op has not been removed

Also decide the symmetric case: register succeeded, upload failed. And what the op queue does in each — retry, dead-letter with bytes preserved (`engine.md:231-235`), or drop.

## Existing machinery

`net::publish::publish` (`net/publish.rs:115`) already implements register-first, floor-plus-one CAS, fanout PUT, and confirm-by-re-resolve — but is only ever called by liveness (`net/liveness.rs:28`) and revival (`net/revival.rs:15`). `PublishRequest.content_cids` (`net/publish.rs:38-53`) exists and every production producer passes an empty vec (`facade.rs:1162` — that is #797).

A new child also needs first-publish-at-sequence-1 semantics (`core.md:207`) plus the parent folder record republished with the new child ref.

FSM1:
## Resolution

**Upload everything, then publish — which registers. Never roll back on failure; retire only on abandonment. One op means all its publishes or none, drained in strict FIFO with a bounded-retry-then-dead-letter valve.**

```
drain(op)
[...]

PR #941, opened 2026-08-01
fix: mark a leaf uploaded before releasing its staged bytes

### LINKED ISSUE #924: engine: a crash between releasing a staged leaf and marking it destroys a successful upload
## Problem

The drain's per-leaf upload step orders its three durable effects `upload → remove → mark` (`crates/engine/src/sync/drain.rs`):

```rust
self.upload_block(&block).await?;
self.staging.remove_staged_bytes(leaf_cid).await.map_err(seam)?;
self.mark_uploaded(&staged.root_cid, index + 1).await?;
```

A crash — or a staging-store failure — between the removal and the mark leaves the mark at `index` while leaf `index` is already gone from staging. On the next pass `upload_mark` returns `index`, the loop reaches leaf `index`, `staged_block` answers `None`, and the hole guard `None if index >= uploaded => Err(CONTENT_LOST)` fires.

`CONTENT_LOST` is `Halt::Permanent(DeadLetterReason::ContentUnrecoverable)`, so the failure valve abandons the op and `release_staged_blocks` discards the rest of the version — **destroying a write whose bytes had in fact uploaded successfully.**

The window is one `remove_staged_bytes` wide per leaf, and it is reachable on any leaf of any version.

## Why the ordering is the wrong way round

The comment on `mark_uploaded` states the intent: "Written after the removal, so the mark never claims more than the store has released." That protects an invariant nothing depends on. The mark is only ever consulted on the `None` arm — a leaf that is still *present* is uploaded again regardless of what the mark says, and re-upload is free (`registerPin` short-circuits an already-pinned CID, #819).

So marking **before** the removal is strictly safer: a crash in the new window leaves the leaf both marked and staged, and the next pass simply re-uploads and re-removes it. No arm of the loop misreads that state, and the "present blocks are a suffix" invariant is untouched.

## Scope

- Swap the two statements so the durable mark is written before the staged bytes are released, and correct the comment on `mark_uploaded` to state the invariant it actually protects.
- A regression test needs staging-store failure injection at `put_staged_bytes` / `remove_staged_bytes` so the interrupted-between-steps state can be reached deterministically — the test fakes have no such hook today, so this slice adds one.
- Confirm `an_evicted_prefix_of_the_block_set_is_loss_not_progress` and `a_hole_in_the_staged_block_suffix_is_loss_and_fails_closed` still hold: a genuinely evicted prefix must stay loss, and only the crash residue may become recoverable.

## Provenance

Found by the `/crypto-privacy-review` gate on #896 while verifying that the new upload-progress emissions did not weaken the suffix invariant. The defect is in code #868 landed and predates #896, which touches neither statement.

## Gate

`Engine Tests`.

## Related

- #869 — owns reconciling the four-way staged-byte lifetime in one place. This ordering bug releases a version's staged bytes under the "proven unrecoverable" rule for a case that is in fact recoverable, so it should land **before** #869 rewrites this loop, or be folded into it. Not a hard blocker in either direction: #869 can proceed on the current ordering, and this fix stands alone.
- #868 — closed; the content write plane that introduced the ordering.

## Depends on

None — the ordering it swaps is in code #868 already landed, and the staging-store failure-injection hook the regression test needs is added by this slice. #869 is not a blocker in either direction (see *Related*).

## Blocks

- #940 — bounds the upload mark count `Drain::upload_mark` parses. This slice makes the mark the sole record of which released leaves are recoverable, so the weight placed on its contents goes up and an unbounded count becomes worth closing.

Part of #655

FSM1:
## Problem

The drain's per-leaf upload step ordered its three durable effects `upload → remove → mark`. A crash — or a staging-store failure — between the removal and the mark leaves the mark at `index` while leaf `index` is already gone from staging. The next pass reads the mark as `index`, reaches leaf `index`, gets `None` from `staged_block`, and the hole guard `None if index >= uploaded => Err(CONTENT_LOST)` fires.

`CONTENT_LOST` is `Halt::Permanent(DeadLetterReason::ContentUnrecoverable)`, so the failure valve abandons the op and `release_staged_blocks` discards the rest of the version — destroying a write whose bytes had in fact uploaded successfully. The window is one `remove_staged_bytes` wide per leaf, and it is reachable on any leaf of any version.

## Change

Two changes, both about the same invariant.

### 1. The mark is written before the staged bytes are released

The ordering argument, which is the substance of this slice:

- The mark is only ever consulted on the `None` arm. A leaf that is still *present* is uploaded again regardless of what the mark says, and re-upload is free — `registerPin` short-circuits an already-pinned CID (#819).
- So the old invariant ("the mark never claims more than the store has released") protected nothing that any reader depends on, while the state it permitted — a released leaf the mark does not cover — is exactly the state the hole guard reads as unrecoverable loss.
- Under the new ordering the residue is a leaf both *marked* and *staged*. The `Some` arm re-uploads and re-removes it. The "still-staged blocks are a suffix" invariant is untouched, because removal still happens only after a confirmed `UploadResult`, in file order.

This is the same rule `crates/desktop-seams` already applies to the op record and its sidecar (hard constraint 5, `staging_store_removal_ordering_leaves_only_a_reclaimable_orphan`): order the durable effects so an interruption leaves a **harmless orphan**, never a dangling reference. A marked-and-staged leaf is the harmless orphan; a released-but-unmarked leaf is the dangling reference.

### 2. The mark only ever rises

Marking first makes "staged at an index *below* the durable mark" a reachable state for the first time, and the `Some` arm wrote `mark_uploaded(index + 1)` unconditionally — so re-uploading such a leaf would **lower** the durable mark below leaves an earlier pass had already released. One further interruption and those genuinely-uploaded leaves are absent and uncovered: `CONTENT_LOST`, valve, version destroyed. The same destruction, one step later.

The reachable source is a release that is *reported done and never persists*: `packages/client/src/seams/stagingStore.ts` flushes a staged write with `handle.flush()` but releases with a bare `dir.removeEntry` and no durability barrier, so "the later flushed mark survives, the earlier unflushed unlink does not" is the crash shape to expect on OPFS. The mark is now written only when it advances, which also saves a durable write per re-uploaded leaf.

Found by the `/crypto-privacy-review` gate on this PR's own diff.

## Tests

All in `crates/engine/tests/write_plane.rs`, gated by `Engine Tests`. The test fakes had no way to reach an interrupted-between-steps state, so this adds injection to `InMemoryStagingStore`: `interrupt_staged_write_after` / `interrupt_staged_removal_after` let `n` calls at a given staging key through, fail the next, and disarm — a one-shot outage at an exact step whose retry then proceeds — and `drop_staged_removal_after` reports a removal done without performing it, the lost-unlink shape.

- **`an_interrupted_leaf_mark_never_costs_the_version_its_uploaded_bytes`** — drops the durable mark at **every leaf in turn**. At each interruption point it asserts the state explicitly: the mark stands exactly at the leaf under test, no dead letter fired, and the op keeps its place at the head of the queue. Then the store returns, the next pass runs, and a second device that only ever saw the network reads the file back byte-for-byte.
- **`a_leaf_left_marked_and_staged_is_re_uploaded_and_released_by_the_next_pass`** — the window the fix opens, at **every leaf in turn**. Asserts the residue directly: the mark covers the leaf **and** the leaf is still staged. The next pass re-uploads it, re-removes it, dead-letters nothing, and leaves only queue bookkeeping.
- **`a_re_uploaded_leaf_never_pulls_the_mark_back_over_leaves_it_covers`** — the regression path for change 2. Leaf 0's release is silently lost while leaves 1 and 2 upload and mark; every pass is stopped at leaf 3. The resumed pass re-uploads leaf 0 and must leave the mark at 3, not pull it back to 1.

Reverting each fix on its own:

```text
# statements swapped back to upload -> remove -> mark
---- an_interrupted_leaf_mark_never_costs_the_version_its_uploaded_bytes ----
panicked at: the resumed pass reads the residue as progress, not as content loss
---- a_leaf_left_marked_and_staged_is_re_uploaded_and_released_by_the_next_pass ----
panicked at: the mark covers the leaf whose release was interrupted
  left: None      right: Some((<root cid>, 1))
test result: FAILED. 45 passed; 2 failed

# high-water guard removed, ordering kept
---- a_re_uploaded_leaf_never_pulls_the_mark_back_over_leaves_it_covers ----
panicked at: re-uploading leaf 0 leaves the mark where it stood: it only ever rises
  left: Some((<root cid>, 1))      right: Some((<root cid>, 3))
test result: FAILED. 47 passed; 1 failed
```

`an_evicted_prefix_of_the_block_set_is_loss_not_progress` and `a_hole_in_the_staged_block_suffix_is_loss_and_fails_closed` both still pass — a genuinely evicted prefix and a genuine hole stay loss. Only the crash residue becomes recoverable.

## Verification

All exit 0 on this branch:

- `cargo fmt --all --check`
- `cargo test -p cipherbox-engine` — 734 passed, 0 failed
- `pnpm -r --if-present run typecheck` and `pnpm -r --if-present run test` — api 177, web 67, all passed
- `pnpm exec eslint .` — clean. No TypeScript surface changed.

Review gates run on this diff: `/security-review` found nothing; `/simplify` returned six cleanups, all folded in — doc trims, and the injectors renamed from `fail_*` to `interrupt_*` so they do not read as the permanently-failing `fail_enqueue_after` convention in the same fake; `/crypto-privacy-review` found the mark-regression defect above, also folded in.

## Deferred

#940 — `upload_mark` bounds the root CID it accepts but not the count, so a corrupt-but-parseable mark can skip every absent leaf and publish a manifest naming blocks nothing holds. Pre-existing, and left out because the repair is a real design choice: reading a corrupt mark as zero destroys a write whose bytes are on the network, while clamping is the maximally trusting reading of bytes already known to be corrupt. #940 depends on this PR and should land after it.

## Note on parallel work

#869 owns reconciling the four-way staged-byte lifetime and will rewrite this loop. This change is scoped to the `Some` arm and the fake's new injection hooks, and touches no shared staging helper, so it should rebase cleanly either way. Per #924 it should land before #869 rewrites the loop, but neither blocks the other.

Closes #924

Part of #655

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when uploading and releasing staged data.
  * Upload progress now remains consistent across retries and recovery scenarios.
  * Previously processed data is safely re-uploaded when needed without losing progress.
  * Fixed interruptions that could leave staged content unreleased or incorrectly marked as complete.
  * Added safeguards to prevent valid uploaded content from being incorrectly discarded.
  * Improved recovery of partially completed uploads across devices.

### DIFF
--- a/crates/engine/src/sync/drain.rs
+++ b/crates/engine/src/sync/drain.rs
@@ -1352,7 +1352,13 @@ where
             match self.staged_block(leaf_cid).await? {
                 Some(block) => {
                     self.upload_block(leaf_cid, &block).await?;
-                    self.mark_uploaded(&staged.root_cid, index + 1).await?;
+                    // A leaf a lost release left staged behind the mark is
+                    // re-uploaded here, and must not drag the mark back down
+                    // over the leaves past it — those are released, so an
+                    // uncovered one reads as loss (#924).
+                    if index + 1 > uploaded {
+                        self.mark_uploaded(&staged.root_cid, index + 1).await?;
+                    }
                     self.staging
                         .remove_staged_bytes(leaf_cid)
                         .await
@@ -1417,12 +1423,11 @@ where
         Ok(<[u8; 4]>::try_from(count).map_or(0, |c| u32::from_be_bytes(c) as usize))
     }
 
-    /// Record that `count` of this version's leaves have uploaded. Written
-    /// *before* the leaf is released, so an interruption between the two leaves
-    /// that leaf both marked and staged and the next pass simply re-uploads and
-    /// re-removes it. The reverse order strands a released leaf behind the mark,
-    /// where the hole guard reads bytes that did upload as unrecoverable loss
-    /// and the valve destroys the version (#924).
+    /// Record that `count` of this version's leaves have uploaded. A high-water
+    /// mark, written *before* the leaf is released: it may over-claim a leaf
+    /// still staged, which the next pass re-uploads, but must never lag or
+    /// regress below one already released — the hole guard would read those
+    /// uploaded bytes as loss (#924).
     async fn mark_uploaded(&self, root_cid: &[u8], count: usize) -> Result<(), Halt> {
         let mut mark = root_cid.to_vec();
         mark.extend_from_slice(&(count as u32).to_be_bytes());
--- a/crates/engine/src/sync/drain.rs
+++ b/crates/engine/src/sync/drain.rs
@@ -1352,11 +1352,11 @@ where
             match self.staged_block(leaf_cid).await? {
                 Some(block) => {
                     self.upload_block(leaf_cid, &block).await?;
+                    self.mark_uploaded(&staged.root_cid, index + 1).await?;
                     self.staging
                         .remove_staged_bytes(leaf_cid)
                         .await
                         .map_err(seam)?;
-                    self.mark_uploaded(&staged.root_cid, index + 1).await?;
                     emit(OpPhase::UploadProgress, blocks(index + 1));
                 }
                 // Absent and not covered by the mark: these bytes were never
@@ -1417,8 +1417,12 @@ where
         Ok(<[u8; 4]>::try_from(count).map_or(0, |c| u32::from_be_bytes(c) as usize))
     }
 
-    /// Record that `count` of this version's leaves have uploaded. Written after
-    /// the removal, so the mark never claims more than the store has released.
+    /// Record that `count` of this version's leaves have uploaded. Written
+    /// *before* the leaf is released, so an interruption between the two leaves
+    /// that leaf both marked and staged and the next pass simply re-uploads and
+    /// re-removes it. The reverse order strands a released leaf behind the mark,
+    /// where the hole guard reads bytes that did upload as unrecoverable loss
+    /// and the valve destroys the version (#924).
     async fn mark_uploaded(&self, root_cid: &[u8], count: usize) -> Result<(), Halt> {
         let mut mark = root_cid.to_vec();
         mark.extend_from_slice(&(count as u32).to_be_bytes());
--- a/crates/engine/src/sync/drain.rs
+++ b/crates/engine/src/sync/drain.rs
@@ -1332,9 +1332,10 @@ where
         })?;
 
         // File order, root last, each leaf removed on its confirmed
-        // `UploadResult` — so the blocks still staged are a **suffix**. An
-        // absence is only progress up to the durable mark this pass keeps: past
-        // it, a missing block is loss, and the version can never be assembled.
+        // `UploadResult` — but a lost release can strand one staged anywhere
+        // below the mark (#924). An absence is only progress up to the durable
+        // mark this pass keeps: past it, a missing block is loss, and the
+        // version can never be assembled.
         let uploaded = self.upload_mark(&staged.root_cid).await?;
         // The root manifest is block zero and goes up last, so the version's
         // whole block count is its leaves plus one.
--- a/crates/engine/src/testkit/fakes/staging_store.rs
+++ b/crates/engine/src/testkit/fakes/staging_store.rs
@@ -14,6 +14,7 @@ struct Inner {
     enqueue_budget: Option<u64>,
     staged_write_budget: Option<(Vec<u8>, u64)>,
     staged_removal_budget: Option<(Vec<u8>, u64)>,
+    dropped_removal_budget: Option<(Vec<u8>, u64)>,
 }
 
 impl Default for Inner {
@@ -27,6 +28,7 @@ impl Default for Inner {
             enqueue_budget: None,
             staged_write_budget: None,
             staged_removal_budget: None,
+            dropped_removal_budget: None,
         }
     }
 }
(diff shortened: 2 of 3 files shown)


FSM1:
### Review response

One finding in this pass, delivered as an outside-diff-range comment in the review body (no inline threads, no other reviewer).

**`crates/engine/src/sync/drain.rs` 1334-1336 — stale "suffix" invariant comment — valid, fixed in d77ddc5**

Confirmed on the branch. `a_re_uploaded_leaf_never_pulls_the_mark_back_over_leaves_it_covers` drops the release of leaf 0 while leaves 1 and 2 are genuinely removed, so the staged residue is provably not a trailing suffix. The claim was true before this PR and this PR is what falsified it.

The correction is narrower than the proposed diff: the comment now drops the false claim and says a lost release can strand a leaf staged anywhere below the mark, keeping the load-bearing half of the sentence (absence is progress only up to the durable mark). The reason a residual leaf is harmless — it is re-uploaded and must not drag the mark back down — already lives at its home, on the `index + 1 > uploaded` guard inside the loop, so restating it here would duplicate it against the repo's state-it-once rule.

Verification: `cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo check --workspace --all-targets`, `cargo test -p cipherbox-engine` (712 tests, 0 failures) all clean. Comment-only change, no behaviour touched.

Issue #796, opened 2026-07-23
engine: surface resolve-tick outcomes as steady-state trust-violation events


FSM1:
The resolve-tick loop (`facade.rs`) currently does `let _ = resolve_and_hold(...)`, discarding the `Resolved.outcome` — including `ResolveOutcome::TrustViolation`. Data stays fail-closed (a forged record is never held or rendered; last-known-good stays pinned), but a persistently forged owner-root record during steady-state polling produces **no event and no log**.

**Acceptance:** wire the staleness-ladder / `AttributableAbuse` surfacing for the tick-loop resolve outcomes, parallel to the `Event::RenewalFailed` surfacing already added for the liveness loop. A persistently-forged steady-state record must raise an event, not vanish.

Surfaced by the #789 security review (LOW/MED, deferred). Part of #752.

FSM1:
Scope note from #917: the focus-window folder refresh (`crates/engine/src/net/focus.rs`) landed a second resolve leg with the same gap. A child-gate rejection on a focused folder — a foreign node id, a foreign scope, a file body behind a folder ref — is fail-closed (last-known-good is pinned, no floor moves) but discarded with no event and no log, exactly like the root leg this issue was filed against.

So the surfacing this issue asks for now has to cover the whole focus chain, not just the owner-root record: the tick resolves the open folder and every ancestor to root through `ChildAdopter`, and `Command::SetFocus` runs the same pass on navigation. A persistently-forged record at any of those names must raise an event.

PR #945, opened 2026-08-01
feat: resolve the focus window's folders below the scope root

### LINKED ISSUE #917: engine: nothing resolves a folder record below the scope root on a device that did not author it

FSM1:
## Problem

Nothing resolved a folder record below the scope root on a device that did not author it, so such a device rendered every subtree empty — indefinitely.

- `spawn_resolve_tick_loop` refreshed the **vault root only**; `refresh_base_from_outcome` to `project_root` lifts direct children only.
- The only other writer into the base below the root was the drain's self-adopt — state this device published itself.
- `Command::SetFocus` fell through to `EngineError::Unimplemented`, which `apps/web/src/engine/snapshotStore.ts` turns into a committed error state on every navigation.
- `ChildAdopter`'s single caller was `read_content`, which is file-only.
- `focus_set` / `FocusWindow` / `FocusTarget::Folder` were pure planning with no caller.

`blueprint/engine.md` "Sync core" specifies both halves — the focus-window tick and the on-access refresh past the staleness threshold. Neither was wired.

## Change

- **`crates/engine/src/net/focus.rs` (new)** — `FolderRefresh` resolves each focus folder's own record cache-first, passes the `ChildAdopter` gate on this device's floors, and merges the gate-passing body with `project_folder`. A gate rejection is fail-closed: last-known-good stands and no floor moves. The pass merges root-ward, so a parent that dropped a child unlinks it before the pass would project into it, and it reads no clock — the caller stamps.
- **`resolve_child` in `net/child.rs`** — the `Adopted` / `Current` / `NoUpdate` + at-floor re-open walk, previously duplicated. Both read paths that descend below the scope root (a file's content read and this refresh) now share it, so neither can drift on which outcome is staleness and which is a fail-closed violation.
- **`Command::SetFocus`** records the focus window and refreshes it immediately, but only for folders past the staleness threshold — navigation is the tick model's second trigger source, and a repeat visit renders state already held. Stamps record *attempts*, so a folder that never gate-passes cannot turn every navigation into a fresh endpoint fan-out; the poll leg still refreshes the window unconditionally, so recovery is automatic.
- **The resolve tick** refreshes the same window every pass, after the root leg and before the drain, so the queue rebases onto the deepest state the pass reconciled.
- **`focus_folders` / `focus_folders_due`** in `sync/tick.rs` — the window's folder targets below the scope root, and the threshold filter. The root rides the vault-pointer leg, never the child gate.
- **`project_folder` reports whether the merge moved the base**, so re-merging the same body every tick does not repaint the host. `Snapshot::link` reports whether it established or raised a link, so the projection no longer predicts it; `project_root` and `refresh_base_from_outcome` forward the report.

No new crypto, no new seam, no TypeScript change — `setFocus` was already plumbed through `crates/wasm` and `packages/client`, and the web store already pulls on success.

## Tests

`crates/engine/tests/write_plane.rs`, gated by **Engine Tests** (`cargo test -p cipherbox-engine`):

- `a_second_device_lists_below_the_scope_root_once_it_focuses_there` — the facade half of #895's round trip. A cold second device lists the depth-2 child with device A's name and node id after setting focus on the depth-1 parent, having listed nothing there before.
- `a_planted_focus_record_never_renders` — fail-closed on all three child-gate bindings the pass rests on. A legitimate concurrent record reconciles first, so the negatives cannot pass on a leg that never ran; then a foreign node id, a foreign scope, and a file body behind a folder ref are each planted at the focused name at a strictly newer sequence, and last-known-good is pinned through all three.
- `an_unreachable_record_plane_leaves_the_focused_folder_rendering` — every endpoint failed, the folder keeps rendering off the cache rather than clearing.
- `navigation_re_resolves_a_folder_only_past_the_staleness_threshold` — a repeat visit inside the threshold renders held state; the same navigation past it reconciles.

Plus `crates/engine/tests/facade.rs::set_focus_records_a_window_with_nothing_to_resolve`, `sync/tick.rs` unit tests for `focus_folders` and `focus_folders_due`, and `sync/project.rs::project_root_reports_a_change_only_when_the_body_moves_the_base`.

**Revert check:** stubbing `focus_folders` to return an empty list fails exactly the four new `write_plane` tests and nothing else.

## Verification

All exit 0:

- `cargo fmt --all`
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test --workspace` — 32 suites green, 0 failures
- `pnpm -r --if-present run typecheck` — api, client, web all clean
- `pnpm -r --if-present run test` — client, api 177, web 67 all pass
- `npx eslint .` — clean

## Review gates

`/simplify`, `/security-review` and `/crypto-privacy-review` were each run on `git diff main...HEAD`. Security and crypto returned no findings. The second commit folds in the real items from all three: the shared `resolve_child` walk, the attempt-stamped on-access damper, `Snapshot::link`'s change report, the root-ward merge order, the extra fail-closed test coverage, and a comment-discipline pass.

One deferral: a child-gate trust violation on the focus leg is fail-closed but unsurfaced to the host, exactly as the root leg is. That is #796's scope; this PR widens what #796 has to cover from one record to the focus chain, and #796 has been updated to say so.

Closes #917

## Summary by CodeRabbit

* **New Features**
  * Focused folders now refresh automatically when stale and immediately after focus changes.
  * Deep-folder views update with newly available content while preserving existing state during refresh failures.
  * Snapshot updates are emitted when refreshed data changes.

* **Bug Fixes**
  * Improved handling of unavailable or rejected child content.
  * Prevented invalid records from replacing valid folder data.
  * Avoided reporting changes when refreshed data is identical.

### DIFF
--- a/crates/engine/src/facade.rs
+++ b/crates/engine/src/facade.rs
@@ -38,10 +38,10 @@ use crate::entropy::Entropy;
 use crate::gate::{GateError, floor};
 use crate::hex::hex_lower;
 use crate::net::{
-    Adopter, ChildAdopter, EolRenewResult, FolderRefresh, FolderRefreshReport, HeldMaterial,
+    Adopter, ChildAdopter, ChildResolveError, EolRenewResult, FolderRefresh, HeldMaterial,
     HeldRecord, HeldRecords, LivenessControl, PublishError, PublishOutcome, RE_PUT_INTERVAL,
     RecordPointerFetch, ResolveOutcome, RootAdopter, eol_renew_pass, keyless_re_put,
-    refresh_base_from_outcome, resolve, resolve_and_hold, run_liveness_loop,
+    refresh_base_from_outcome, resolve_and_hold, resolve_child, run_liveness_loop,
 };
 use crate::profile::SyncTimingProfile;
 use crate::rotation::derive_write_name;
@@ -62,7 +62,7 @@ pub use crate::sync::rebase::DeadLetterReason;
 use crate::sync::record::{RecordReader, RecordSeal};
 use crate::sync::staging::stage_op;
 use crate::sync::staleness::{Connectivity, classify};
-use crate::sync::tick::{FocusWindow, focus_folders, on_access_refresh_due};
+use crate::sync::tick::{FocusWindow, focus_folders, focus_folders_due};
 
 /// The stable 16-byte node identifier (`id16`, blueprint/core.md). Public,
 /// non-secret, and location-independent — routes and commands key on it,
@@ -880,16 +880,17 @@ fn node_attrs(meta: &NodeMeta) -> NodeAttrs {
     }
 }
 
-/// Stamp every folder a focus pass merged with the caller's clock reading, so
-/// the on-access threshold measures from the last gate-passing merge. A folder
-/// the pass could not merge keeps its old stamp and stays due.
+/// Stamp every folder a focus pass attempted against the caller's clock
+/// reading. Attempts, not merges: an unresolvable folder must not turn every
+/// navigation into a fresh endpoint fan-out, and the poll leg refreshes the
+/// window regardless, so recovery stays automatic.
 fn stamp_focus_refreshed(
     stamps: &RefCell<BTreeMap<NodeId, UnixMillis>>,
-    report: &FolderRefreshReport,
+    folders: &[NodeId],
     now: UnixMillis,
 ) {
     let mut stamps = stamps.borrow_mut();
-    for folder in &report.refreshed {
+    for folder in folders {
         stamps.insert(*folder, now);
     }
 }
@@ -1131,9 +1132,9 @@ pub struct Engine<T: SeamTypes> {
     /// open, whose record and whole ancestor chain every resolve tick refreshes.
     /// Shared with the tick loop, which reads it on each pass.
     focus: Rc<RefCell<FocusWindow>>,
-    /// When each focus folder last merged a gate-passing body, so a navigation
-    /// inside the staleness threshold renders state already held instead of
-    /// re-resolving (blueprint/engine.md: refresh on access past the threshold).
+    /// When each focus folder was last refreshed, so a navigation inside the
+    /// staleness threshold renders state already held instead of re-probing the
+    /// record plane (blueprint/engine.md: refresh on access past the threshold).
     focus_refreshed: Rc<RefCell<BTreeMap<NodeId, UnixMillis>>>,
     /// Retained dead-lettered ops. Feeds [`SnapshotView`]'s dead-letter surface
     /// (#33 D6: dead letters are retained, never silent).
@@ -1592,7 +1593,7 @@ impl<T: SeamTypes> Engine<T> {
                 if let Some(read_seed) = &read_seed
                     && !open.is_empty()
                 {
-                    let report = FolderRefresh {
+                    let changed = FolderRefresh {
                         transport: &transport,
                         snapshot_cache: &snapshot_cache,
                         http: &http,
@@ -1604,8 +1605,8 @@ impl<T: SeamTypes> Engine<T> {
                     }
                     .run(&open)
                     .await;
-                    stamp_focus_refreshed(&focus_refreshed, &report, scheduler.now());
-                    if report.changed {
+                    stamp_focus_refreshed(&focus_refreshed, &open, scheduler.now());
+                    if changed {
                         let _ = events.unbounded_send(Event::SnapshotUpdated);
                     }
                 }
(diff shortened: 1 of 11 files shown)


FSM1:
Addressed the review-body nitpick. No inline threads were opened on this pass, so replying here.

**`crates/engine/tests/write_plane.rs` 1994-2029 — reuse the fixture in the depth-2 test — accepted, with a correction.**

The duplication is real: `a_create_below_the_scope_root_is_adoptable_by_a_second_device` repeated the fixture's create-and-boot prologue verbatim.

The suggestion as written would not have compiled, though. It proposed reusing "the helper's returned world, blocks, second engine, tasks, photos, and deep node", but that test also needs device B's own seams — `bob.http`, `bob.floor_store`, `bob.record_store`, `bob.snapshot_cache` — to build the `ChildAdopter` and call `resolve`. The fixture did not return the `FakeDevice`, and it cannot be re-derived at the call site: `FakeWorld::device` mints fresh device-local stores on every call, so a second `world.device(b"alice-second-device")` would hand back a cold floor store and cache that the booted engine never touched, silently changing what the test asserts.

Fixed in 7f7524b4c by returning a named `DeepCreate` struct rather than widening the 6-tuple to 7. The struct also keeps the call sites self-documenting — named fields plus `..` in place of six positional `_world, _blocks, _tasks_b` placeholders.

The record-plane test's doc comment claimed "the facade has no descent below the scope root yet", which this PR makes false; it now cross-references its facade half instead.

Verified: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo check --workspace --all-targets`, `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown --all-targets`, and `cargo test --workspace` all exit 0, with no test failures.

Issue #646, opened 2026-07-20
desktop: implement the desktop seam set against the conformance kits


FSM1:
## Scope

The desktop seam implementations under <data_local_dir>/cipherbox/<accountId>/, ciphertext-only at rest: FloorStore (fsync-barriered files), StagingStore (generalized v1 write journal: one durable record per op, sidecar staged ciphertext, json-before-bin removal, orphan-sidecar GC), SnapshotCache (sealed cache files), CredentialStore (keyring; refresh token + last-account id only), Http/RecordTransport (reqwest), Scheduler (Tokio).
Does NOT construct the engine at login or touch the webview.

## Blueprint anchors

`blueprint/desktop.md`: Engine wiring

## Gate

Seam conformance kits for the desktop implementations inside the cargo workspace tests PR gate.

## Depends on

- #645 (desktop: scaffold the apps/desktop Tauri shell and version surface)
- #623 (engine: scaffold seams, fakes, conformance kits, and the facade)

PR #952, opened 2026-08-01
fix: barrier desktop store unlink ordering where directories cannot be fsynced

### LINKED ISSUE #665: desktop: harden StagingStore unlink-ordering durability on Windows
## Summary

The desktop `FileStagingStore` (`crates/desktop-seams/src/staging_store.rs`) relies on the op-record-before-sidecar removal ordering as a crash-consistency property: the engine removes the op record first, then its staged-ciphertext sidecar, so a crash can only ever orphan a sidecar (harmless, GC'd) — never leave an op record pointing at a sidecar that is already gone.

On Unix this is enforced by `remove_file_durable`, which fsyncs the parent directory after each unlink (`fs_util.rs::fsync_dir`), making the op-record removal durable before the sidecar removal is even attempted.

On **Windows**, `fsync_dir` is a deliberate no-op (`#[cfg(not(unix))]` in `fs_util.rs`) — Windows has no directory fsync. `atomic_write`'s durability is still covered there by `MOVEFILE_WRITE_THROUGH` + NTFS journaling, but **unlink ordering is not barriered**. NTFS metadata journaling makes the volume *consistent* after a crash, but does not guarantee the app-visible ordering that the op-record unlink is persisted before the sidecar unlink. The dangerous inverse (sidecar gone, op record still present) is therefore theoretically possible on Windows.

## Impact

- Platform: Windows only. Unix (macOS/Linux) is fully barriered and unaffected.
- Severity: low — the exposure is a crash in the narrow window between two specific unlinks, and the failure mode is a dead-lettered op, not data loss. But Windows is a real build and run target: the WinFsp adapter is not CI-only — it executes on actual Windows machines — so this is a genuine durability gap on a shipping platform, not a theoretical CI-only concern.
- If the inverse state occurs, a replayed op would reference a staging key whose bytes are gone, surfacing as a failed/dead-lettered op rather than data loss.

## Suggested direction

- Add a Windows-durable unlink barrier for the StagingStore completion path (e.g. FlushFileBuffers on a handle to the containing directory where supported, or a tombstone/two-phase completion marker that survives a metadata reorder), and assert the ordering in a Windows-gated test.
- Alternatively, document the Windows guarantee as journaling-consistency-only and gate the stronger claim behind Unix.

Reference: surfaced during the ship review of the desktop seam set (`#646`). The current code comment in `fs_util.rs` should be revisited so it does not over-claim cross-platform crash-ordering.

## Depends on

None — #646 landed both files this touches (`crates/desktop-seams/src/staging_store.rs`, `fs_util.rs`) and the Windows CI leg that would gate the new ordering test already runs.

Part of #655

FSM1:
## Problem

`FileStagingStore` documents removal ordering as a correctness property (desktop-seams hard constraint 5): the op record goes first, its staged-ciphertext sidecar second, so an interruption can only orphan a sidecar — never leave an op record naming a sidecar that is already gone. `remove_file_durable` enforces that by fsyncing the parent directory after each unlink.

`fsync_dir` was `Ok(())` on non-Unix. Windows has no directory fsync, so on that platform the ordering was never barriered at all and the store silently reported a durability it had not obtained.

**The over-claim is wider than the issue recorded.** #665 assumed `atomic_write` was still covered on Windows by `MOVEFILE_WRITE_THROUGH`, per the comment in `fs_util.rs`. That comment is wrong. Rust std's Windows `rename` is `MoveFileExW(old, new, MOVEFILE_REPLACE_EXISTING)` — write-through is not among the flags (`library/std/src/sys/fs/windows.rs`, `fn rename`, checked against the 1.88 sources this workspace builds with). So the rename was unbarriered too, and the fix has to cover both barriers, not just unlinks.

## Change

`fsync_dir` on non-Unix now calls `metadata_log_barrier`: write a byte to a fresh temp file in the same directory and `sync_all` it. NTFS journals metadata to a per-volume write-ahead log flushed as an LSN-ordered prefix, so flushing a transaction issued *after* a directory-entry change also persists that change. The temp carries the existing `.cbtmp.` prefix, so it is invisible to `list_file_names` and a crash before its removal leaves debris `ensure_dir` already sweeps — the harmless orphan, in line with the same principle the store's ordering rule serves.

The barrier is plain `std::fs`; the crate is `#![forbid(unsafe_code)]` and no Win32 call was added. It is compiled under `cfg(any(not(unix), test))`, so the Windows algorithm is exercised by the Unix CI legs too rather than only by the Windows one.

Also in the diff:

- `atomic_write` and `metadata_log_barrier` share a `write_synced_temp` helper instead of two copies of create/write/`sync_all`.
- A barrier failure in `remove_file_durable` is labelled `unlink barrier: …`, so the new Windows-only failure mode (the barrier's `File::create` hitting `ENOSPC` or a read-only directory after the unlink already succeeded) is diagnosable rather than surfacing as a bare `staging_store remove_op` error.
- The over-claiming comments in `fs_util.rs` and on `FileStagingStore` are corrected; the store's doc now cross-references `fs_util::fsync_dir` rather than restating the platform mechanism.

## Dependencies

The issue carries no dependency statement. Established from its body and the surrounding desktop slices:

- **No blocking prerequisite.** `crates/desktop-seams` is a leaf adapter crate. The change is confined to its private `fs_util` module and touches no seam trait, no engine code, and no wire format. It is independently mergeable.
- **Depends on #646 only historically** — the seam set whose ship review surfaced this. That work is landed.
- **Not fixable a layer down.** `FileStagingStore` stores op bytes verbatim and never parses them, so it cannot know which sidecar an op references and cannot repair the dangerous state at reopen. An ordering barrier is the only enforcement available at this layer.

### Relationship to #941 and #950

Both are open against the **engine-side** staged-block lifecycle; this is the **desktop seam** underneath them. No overlap in files, and the rules do not contradict:

- **#941** (issue #924, `crates/engine/sync/drain.rs`) reorders mark-before-release for a leaf so an interruption leaves a marked-and-staged leaf rather than a released-but-unmarked one, and cites this store's hard constraint 5 by name as the rule it is applying. This PR makes that same rule actually hold on Windows, where the barrier it depends on was absent. Strictly reinforcing.
- **#950** (issue #869) adds cancel plus orphan-staged-block GC. Its GC is the reclaimer for exactly the orphan this store's ordering is designed to leave behind, and it decides which blocks are collectible from the queued op records — which this change makes more trustworthy on Windows. No shared file; the `staging_store_removal_ordering_leaves_only_a_reclaimable_orphan` test name #941's body references is preserved.

## Tests

`crates/desktop-seams/src/fs_util.rs`:

- `metadata_log_barrier_fails_closed_when_it_cannot_be_established` — **the revert guard.** Restoring the old `Ok(())` body makes it fail on every platform, verified locally. It pins that the barrier is a real operation with real failure modes, so an unbarriered removal is an error rather than a silent fast path.
- `metadata_log_barrier_leaves_the_directory_as_it_found_it` — successive barriers neither collide on a temp name nor accumulate debris, and do not disturb neighbouring files. This one is a residue guard, not a revert guard: it passes against the old no-op too.

`crates/desktop-seams/tests/conformance.rs`:

- `staging_store_removal_ordering_leaves_only_a_reclaimable_orphan` now walks all four interruption points of one op's life — bytes staged before the op is journaled, op journaled, op record removed before the sidecar, sidecar reclaimed — reopening the store at each and asserting exactly what survived. A `Survivors` precondition rejects any kill point that expects an op record without its sidecar, so the forbidden state cannot be encoded into the table.

## Verification

All exit 0 on macOS (aarch64, Rust 1.88): `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo check --workspace --all-targets`, `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown --all-targets`, `cargo test -p cipherbox-desktop-seams` in both debug and `--release`, `pnpm -r --if-present run typecheck`, `pnpm -r --if-present run test`, `eslint .`.

Beyond that, the non-Unix path was forced on locally — `fsync_dir` temporarily rewired so `metadata_log_barrier` was the only implementation — and the whole `cipherbox-desktop-seams` suite passed against it. That proves the Windows code path compiles and is semantically correct as a barrier-shaped operation.

**What only the Windows CI leg can confirm**, stated plainly:

- That the `#[cfg(not(unix))]` arm compiles and links on `x86_64-pc-windows-msvc`. Cross-checking from macOS is not possible here: `cargo check --target x86_64-pc-windows-msvc` fails in `blake3`'s build script for want of `ml64.exe`. The authoritative check is the `Cargo Check & Test (Windows)` job in `ci.yml`, which runs `cargo test --workspace` on `windows-latest` and therefore executes both new tests against the real production path.
- That the barrier behaves on a real NTFS volume — file creation in the store directories, and the temp's removal, under Windows sharing semantics and any AV or indexer holding handles.

**What no CI leg can confirm**, and I am not claiming: that the NTFS LSN-ordered-log-prefix premise actually delivers the ordering. Proving it needs power-loss or crash-injection testing on real Windows hardware, which is out of reach of this repo's suites. The tests here assert the barrier's shape and its fail-closed behaviour, not the durability guarantee itself. This is a strict improvement over the previous no-op regardless — the old code obtained no ordering under any model.

Desktop E2E is not required for this change; no FUSE or shell surface is touched.

**Windows cost, accepted deliberately.** Every `atomic_write` and every `remove_file_durable` on Windows now pays an extra file create + write + `FlushFileBuffers` + unlink. Scoping the barrier to unlinks only was considered and rejected: with `MOVEFILE_WRITE_THROUGH` absent from std's `rename`, the write path needs it too. Batching the barrier across the delete loops in `snapshot_cache` and `floor_store`, and skipping it in `ensure_dir` when the sweep removed nothing, are real wins but sit outside this diff.

## Review gates

- **`/security-review`** — no HIGH or MEDIUM findings. The diff adds no untrusted input, no crypto, no network surface and no secret handling; the barrier file carries a single NUL byte and never sees key or sealed material. Failure propagates fail-closed, so a failed barrier fails the removal instead of letting the caller proceed to the sidecar.
- **`/simplify`** — findings folded in: the duplicated temp-write block extracted to `write_synced_temp`; the barrier doc cut from 13 lines to 6; the platform mechanism no longer restated on `FileStagingStore`; a tautological assertion in the test replaced with a real precondition on the kill-point table; positional bools replaced with a named `Survivors` struct; absent-versus-corrupted sidecar states no longer conflated. One finding was rejected on evidence — scoping the barrier away from `atomic_write` rests on the `MOVEFILE_WRITE_THROUGH` claim, which the std sources disprove. One accepted change was then reverted: barriering the already-absent `remove_file_durable` path closes a retry-after-barrier-failure window, but costs a directory fsync per already-removed leaf on the per-block GC path, which is not a trade worth making here.
- **`/crypto-privacy-review`** — not applicable; the diff touches no key material and no sealed bytes.

Closes #665

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved durability when saving, updating, and removing staged data, helping ensure changes persist correctly across interruptions or system failures.
  * Strengthened cleanup behavior so incomplete operations and leftover temporary data are handled safely.
  * Added safeguards to detect and report situations where durable filesystem updates cannot be completed.

* **Tests**
  * Expanded coverage for interrupted staging operations, recovery scenarios, orphaned data, and final cleanup states.

### DIFF
--- a/crates/desktop-seams/src/fs_util.rs
+++ b/crates/desktop-seams/src/fs_util.rs
@@ -63,7 +63,7 @@ pub(crate) fn atomic_write(path: &Path, bytes: &[u8]) -> io::Result<()> {
             return Err(err);
         }
     }
-    fsync_dir(dir)
+    fsync_dir(dir).map_err(|err| io::Error::new(err.kind(), format!("write barrier: {err}")))
 }
 
 /// Durably removes a file. Idempotent: a missing file is success. The
--- a/crates/desktop-seams/src/fs_util.rs
+++ b/crates/desktop-seams/src/fs_util.rs
@@ -48,22 +48,14 @@ pub(crate) fn ensure_dir(dir: &Path) -> io::Result<()> {
 /// Durably writes `bytes` to `path`, replacing any existing file atomically.
 ///
 /// Barrier order: write a fresh temp file in the same directory, `sync_all`
-/// its contents, `rename` it over the target (atomic replace on Unix and on
-/// Windows — Rust's `fs::rename` maps to `MoveFileExW` with
-/// `MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH`), then fsync the
-/// directory so the rename itself is durable. A crash can only ever leave
-/// the old value or the new value — never a torn one.
+/// its contents, `rename` it over the target (an atomic replace on both Unix
+/// and Windows), then [`fsync_dir`] so the rename itself is durable. A crash
+/// can only ever leave the old value or the new value — never a torn one.
 pub(crate) fn atomic_write(path: &Path, bytes: &[u8]) -> io::Result<()> {
     let dir = path
         .parent()
         .ok_or_else(|| io::Error::new(io::ErrorKind::InvalidInput, "path has no parent"))?;
-    let tmp = temp_path(dir);
-    // Scope the handle so it is closed before the rename.
-    {
-        let mut file = File::create(&tmp)?;
-        file.write_all(bytes)?;
-        file.sync_all()?;
-    }
+    let tmp = write_synced_temp(dir, bytes)?;
     match fs::rename(&tmp, path) {
         Ok(()) => {}
         Err(err) => {
@@ -75,9 +67,9 @@ pub(crate) fn atomic_write(path: &Path, bytes: &[u8]) -> io::Result<()> {
 }
 
 /// Durably removes a file. Idempotent: a missing file is success. The
-/// removal is barriered with a directory fsync so an ordered caller (e.g.
-/// the StagingStore's op-before-sidecar removal) can rely on it having hit
-/// the platter before the next removal begins.
+/// removal is followed by [`fsync_dir`] so an ordered caller (e.g. the
+/// StagingStore's op-before-sidecar removal) can rely on it having hit the
+/// platter before the next removal begins.
 pub(crate) fn remove_file_durable(path: &Path) -> io::Result<()> {
     match fs::remove_file(path) {
         Ok(()) => {}
@@ -85,7 +77,8 @@ pub(crate) fn remove_file_durable(path: &Path) -> io::Result<()> {
         Err(err) => return Err(err),
     }
     if let Some(dir) = path.parent() {
-        fsync_dir(dir)?;
+        fsync_dir(dir)
+            .map_err(|err| io::Error::new(err.kind(), format!("unlink barrier: {err}")))?;
     }
     Ok(())
 }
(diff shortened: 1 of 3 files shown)


FSM1:
Review of `649df0b9d` triaged — one nitpick, body-only, no threads to resolve.

### `fs_util.rs:70-84` — label the write-side barrier failure

**Applied** in `879803837`.

The asymmetry is real and this PR introduced it: adding `"unlink barrier: {err}"` to `remove_file_durable` left `atomic_write`'s identical `fsync_dir(dir)` failure unlabelled, so after this change a barrier failure on write was harder to diagnose than one on unlink — for two barriers protecting the same crash-consistency guarantee.

```rust
fsync_dir(dir).map_err(|err| io::Error::new(err.kind(), format!("write barrier: {err}")))
```

Gates on the amended branch: `cargo fmt --all --check` clean, `cargo clippy -p cipherbox-desktop-seams --all-targets -- -D warnings` clean, `cargo test -p cipherbox-desktop-seams` 15 passed / 13 passed / 0 failed, 2 pre-existing `real_keyring_*` ignored.

Issue #956, opened 2026-08-01
core: extend the transient-tree scrub guard to the write-body, section, and mailbox codecs


FSM1:
## Context

Raised by the `/crypto-privacy-review` gate on #936. That slice established the terminal-owner doctrine for transient `Value` trees and gave the crate one unified guard pair (`crates/core/src/codec/scrub.rs`: `Scrub`, `ScrubOnDrop`, `ScrubOwned`), then applied it to the paths it touched:

- `decode_read_body` / `encode_read_body` (`crates/core/src/seal/body.rs`)
- all four grant payload codecs (`crates/core/src/seal/grant.rs`)
- `crates/core/src/seal/content_key.rs`
- the decoder's own array/map accumulators (`crates/core/src/codec/decode.rs`)

Three codecs in the same crate were not brought along, so they still drop their transient tree un-wiped.

## Gap

| File | Function | What the tree carries |
| ---- | -------- | --------------------- |
| `crates/core/src/seal/write_body.rs` | `decode_write_body`, `encode_write_body` | recipient secp256k1 identity public keys and blinded grant tags; child-scope `ipnsName`s |
| `crates/core/src/seal/section.rs` | the grant-section decode/encode | the signed blob set |
| `crates/core/src/payload/mailbox.rs` | `inner_map` | a verbatim `Vec<u8>` copy of the sealed payload |

None of these is key material, so this is lower severity than #936's gaps — it is pseudonymous identifiers, already-sealed bytes, and privacy-relevant metadata stranded in freed heap. But #936's own justification for wiping `Value::Text` (a filename is user-private metadata in a zero-knowledge system, `crates/core/src/codec/value.rs`) covers `ipnsName`s and recipient keys just as squarely, so leaving these three inconsistent is a doctrine gap rather than a deliberate carve-out.

## Fix

Wrap each transient tree in `ScrubOnDrop` (encode side) or `ScrubOwned` (decode side), matching the shape `crates/core/src/seal/grant.rs` uses. The guards already exist and are `pub(crate)`, so this is mechanical. Cover it the way #936 does — a test that the guard wipes what it guards, asserting behavior rather than source text.

## Why not in #936

#936 is a hygiene slice that already touches 35 files across `crates/core` and `crates/engine`. Extending it to three more codecs would grow the diff without sharing a review surface: these are separate wire formats with their own KAT families, and each wants its round-trip re-checked on its own.

## Depends on

[...]

Issue #937, opened 2026-07-31
core: redact secret-bearing renderings of Value, Map and preserved unknowns


FSM1:
## Context

Found by all three review gates on #902 (PR #935). That PR gave `codec::UnknownFields` a redacting `Debug` per security rule 2. All three reviewers independently called the boundary a half-measure: the redaction covers one of several ways the same bytes reach a log line.

## Gap

`crates/core/src/codec/value.rs`:

- `Value` derives `Debug` (`:16`) — `Value::Bytes` renders every byte. `Value::Text` too, and `Value::zeroize_bytes`'s own doc (`:132-134`) argues text is user-private in a ZK system.
- `Map` derives `Debug` (`:170`), so a decoded known-field map prints `Map { entries: [("contentKey", Bytes([171, 171, …]))] }`.
- `impl Display for Value` (`:253-267`) renders `Bytes` as **hex**. Its only intended consumer is the KAT `diag` vectors, but any `{}` in a log line leaks.

`crates/core/src/seal/body.rs:193-203`: `impl Debug for Version` redacts `content_key` and then prints `.field("unknown", &self.unknown)` in full, one line below — the preserved unknowns are exactly the bytes #902 argues are secret-bearing.

`crates/core/src/seal/body.rs`: **`ChildRef` derives `Debug`**, so `{:?}` on a `ReadBody::Folder` renders every child's **plaintext filename** and `ipnsName`. Found while implementing #936 (PR #957) and outside that slice's scope. This is not key material, but in a zero-knowledge system the filename set is exactly what the server must never see, so rendering it into a log line or a panic message is the same class of leak as rule 2 covers. Scope widened to include it 2026-08-01.

Sweep the rest of the seal layer for the same shape rather than fixing these three by name: any type that derives `Debug` and holds decoded user plaintext or key material is in scope.

## Suggested direction

- A hand-written `Debug for Value` rendering `Bytes(<n bytes redacted>)` / `Text(<n chars redacted>)`. That covers `Map`, `Version`, and every downstream derive in one place, and is what makes the `UnknownFields` impl mean something rather than being a speed bump.
- Consider moving the diagnostic renderer off `Display` to an explicit `to_diag()`, so no accidental `{}` can leak. The KAT vectors are the only consumer; this is a mechanical call-site change.
- Redact `Version`'s `unknown` field.
- Redact `ChildRef`'s filename and `ipnsName`, and audit sibling seal-layer types that derive `Debug` over decoded plaintext.

[...]

Issue #902, opened 2026-07-29
core: give callers a way to wipe secret-bearing UnknownFields and Map values


FSM1:
## Context

Spotted by the `/crypto-privacy-review` gate on #901, which closed the encode-realloc hygiene gap in `encode_map_partial` (#893). That fix keeps secret bytes to one exactly-sized buffer inside the encoder. The gap below is adjacent, pre-existing, and deliberately out of scope there.

## Gap

The partial codec hands out secret-bearing values that an out-of-crate caller has **no way to wipe**.

1. **`UnknownFields` has no zeroize path at all.** `decode_map_partial` is `pub` and re-exported from `crates/core/src/codec/mod.rs`, so a caller outside `crates/core` receives an `UnknownFields` whose `entries` hold raw canonical value bytes. `Value::zeroize_bytes` and the corresponding `Map` path are `pub(crate)` (`crates/core/src/codec/value.rs`), so the caller cannot reach them. The `UnknownFields` vectors are simply dropped un-zeroized.

2. **`Map::insert` returns the replaced `Value`** (`crates/core/src/codec/value.rs`). The rewrite flow is `known.insert("contentKey", new_key);` — ignoring the return value is the ergonomic default, and it drops the previous secret un-zeroized. Same leak class as #893, on the caller's side of the seam.

## Exposure today

**Latent.** There are currently no callers of `decode_map_partial` / `encode_map_partial` outside `crates/core` — only `crates/core/tests/`, `crates/core/examples/kat_gen.rs`, and the `mod.rs` re-export. This bites the moment a slice wires the unknown-field-preserving re-encode path into the seal/grant surfaces, which is the same trigger #893 called out: that is the point these types start carrying inline content keys and grant/owner/history seed material across the crate boundary.

## Suggested direction

Consistent with AGENTS.md security rule 7 — zeroize at the terminal owner, never in a callee on a caller-owned buffer — the fix is to give the terminal owner the ability it currently lacks:

- `impl Drop for UnknownFields` that zeroizes each preserved raw vector, or a public explicit wipe.
- A public wipe on `Map`, so a caller holding decoded known fields can clear them.
- For `Map::insert`, decide deliberately: either `#[must_use]` on the returned old value so ignoring it is a lint, or an `insert` that zeroizes the value it replaces (and a separate `replace` for callers that want it back).

Land this **before** the first out-of-crate caller feeds secret-bearing fields through the partial codec.

## Dependencies

[...]

PR #957, opened 2026-08-01
fix: wipe the decoder in-flight tree and the seal layer preserved unknowns

### LINKED ISSUE #936: core: wipe the decoder in-flight tree and the seal layer preserved unknowns
## Context

Found by the `/security-review` and `/crypto-privacy-review` gates on #902 (PR #935). #902 closed the wipe seam on `codec::UnknownFields` and `codec::Map`. Two adjacent surfaces in the same crate still drop verbatim copies of secret bytes un-zeroized, and both are on paths that ship today — unlike the partial codec #902 hardened, which has no out-of-crate caller yet.

## Gap 1 — the decoder strands its in-flight tree on a reject

`crates/core/src/codec/decode.rs:167-170` and `:174-179` accumulate into a local before propagating:

```rust
let mut items = Vec::new();
for _ in 0..head.arg {
    items.push(self.read_value(depth + 1)?);   // Err drops `items` un-wiped
}
```

Given `{"k": [<32B secret>, <32B secret>, <malformed>]}`, the two already-decoded `Value::Bytes` are dropped un-zeroized inside the decoder, below any guard a caller can install. This is under the full `decode()` that `crates/core/src/seal/` uses, so it is live on every sealed-body and grant open, not only the partial codec.

Fix: wipe the accumulator before propagating in both arms, or a drop guard over `items` / the map entries. One change covers `decode()` and `decode_map_partial`.

## Gap 2 — `seal/body.rs` preserved unknowns outlive their scrub guard

`collect_unknown` (`crates/core/src/seal/body.rs:473-479`) **clones** values out of the decoded tree:

```rust
.map(|(k, v)| (k.clone(), v.clone()))
```

into `pub unknown: Vec<(String, Value)>` on `ChildRef` (`:82`), `Version` (`:137`) and `ReadBody` (`:232`, `:238`). Those clones outlive the `ScrubOwned` guard's wipe and are never scrubbed. This is the *shipping* preserve-unknowns carrier — the seal layer does not route through `codec::UnknownFields`.

Fix: give these the same treatment (a `Zeroizing`-backed carrier with a wipe), or route the seal layer through `codec::UnknownFields` so one implementation carries the property.

## Why not in #902

Gap 1 is in `decode.rs`, which #902 does not touch, and changes behaviour for every `decode()` caller including the KAT paths. Gap 2 is a public field-type change on `ChildRef`/`Version`/`ReadBody` with engine-side consumers. Both want their own review pass.

## Dependencies

Depends on #902 (PR #935) — that PR establishes the terminal-owner wipe shape this extends to the seal layer and the decoder.

## Blocks

- #956 — extends the same `Scrub` / `ScrubOnDrop` / `ScrubOwned` guard pair to the write-body, grant-section and mailbox codecs. That slice reuses the guards this one lands rather than adding a second mechanism, so it cannot start until they exist.

Part of #655

## Gap 3 — a wiped `Map` re-encodes to silently-empty fields

Raised by CodeRabbit on PR #935 and deferred there as out of scope. **Lower priority than gaps 1 and 2, and a different class:** this is misuse resistance, not a fail-closed hole. Recorded here so the wipe seam is hardened in one pass.

PR #935 promotes `Map::zeroize_bytes` from `pub(crate)` to `pub`. Nothing marks the map afterwards, so both encoders still accept it:

```text
{"contentKey": h'ab…ab' (32B), "name": "secret-file.txt", "v": 2}
  known.zeroize_bytes();
  encode(&Value::Map(known))  -> a3 61 76 02 64 6e616d65 60 6a 636f6e74656e744b6579 40
  decode(that)                -> Ok({"v": 2, "name": "", "contentKey": h''})
```

Identical under `cargo test --release`. No secret survives into the output — the wipe works; the hazard is the opposite direction. A caller that wipes and then re-encodes publishes a structurally valid record whose fields have been emptied.

**Why it is not an AGENTS.md rule 8 violation.** Rule 8 asks for an encode-side `Err` wherever the decoder hard-rejects what the encoder can emit. Here the decoder *accepts* the output above: `h''` and `""` are canonical det-CBOR. The one place a wipe does produce decoder-rejected bytes — an empty preserved raw value, which would splice a key with nothing after it — already has its release-active guard and test in #935 (`crates/core/src/codec/fields.rs:204-206`, `merged_encode_rejects_an_empty_preserved_value`).

**Why the existing rustdoc is not quite enough.** `crates/core/src/codec/value.rs:238-239` states the obligation:

> Terminal: a wiped map still encodes, to zero-length fields a schema decoder rejects, so nothing may re-encode from it afterwards.

The "a schema decoder rejects" half only holds for fixed-length fields. `bytes_fixed::<N>` (`crates/core/src/seal/body.rs:455`) does reject `h''` with `InvalidFieldLength`, but the variable-length reads next to it do not — `ChildRef::from_value` takes `name` via `as_text()` and `ipnsName` via `as_bytes()` with no length floor (`crates/core/src/seal/body.rs:90-91`). A wiped `ChildRef` map would re-encode and re-decode clean with an empty name and an empty `ipnsName`. So the failure mode is silent data loss, not a loud reject.

**Update:** the rustdoc half of this is already fixed on #935 (`f8d2bb8e9`) — it now reads "Fixed-length fields then fail their schema decode, but variable-length ones round-trip empty". Only the enforcement below is still open; do not re-do the doc.

**Not currently reachable.** `encode_map_partial` / `decode_map_partial` have no caller in `crates/core/src` or `crates/engine` — only tests, `kat_manifest.rs` and the `kat_gen` example. Every non-test `Map::zeroize_bytes` call site wipes and then drops (`crates/core/src/seal/body.rs:385,404`, `crates/core/src/codec/fields.rs:115,131`), and `encode_settings_body` encodes before it wipes (`crates/engine/src/settings.rs:392-393`).

**Fix sketch.** A wiped flag on `Map` that `zeroize_bytes` sets and that `write_value` / `count_value` / `encode_map_partial` reject on before emitting a byte, plus a test that runs in a release build. It is a design change to the `Map` type touching every encoder arm, which is why it wants its own pass rather than a review-response commit on #935. Worth deciding alongside gap 1, since a decoder that wipes its in-flight accumulator would be setting that same flag on trees it is about to discard.

FSM1:
## Problem

#902 closed the wipe seam on `codec::UnknownFields` and `codec::Map`. Three adjacent surfaces in the same crate still dropped verbatim copies of secret bytes un-zeroized, and unlike the partial codec #902 hardened, all three are on paths that ship today.

1. **The decoder stranded its in-flight tree on a reject.** `read_value`'s array and map arms accumulated into a bare local, so `{"k": [<32B secret>, <32B secret>, <malformed>]}` dropped the two already-decoded `Value::Bytes` un-wiped inside the decoder — below any guard a caller can install, on every sealed-body and grant open. The top-level `expect_end` reject had the same shape: a whole body decoded, then dropped intact on trailing bytes.
2. **The seal layer's preserved unknowns outlived their scrub guard.** `collect_unknown` *clones* values out of the decoded tree into `pub unknown: Vec<(String, Value)>`. Those clones survive `ScrubOwned`'s wipe and were never scrubbed. This is the shipping preserve-unknowns carrier — the seal layer does not route through `codec::UnknownFields`.
3. **A wiped `Map` re-encoded to silently-empty fields.** Nothing marked the map after `zeroize_bytes`, so both encoders still accepted it. `bytes_fixed::<N>` rejects `h''` afterwards, but the variable-length reads next to it do not — a wiped `ChildRef` round-trips clean with an empty `name` and an empty `ipnsName`. Silent data loss, not a loud reject.

## Change

- **`crates/core/src/codec/scrub.rs`** (new) — one `Scrub` trait plus `ScrubOnDrop` (borrowed, `mem::forget` to disarm) and `ScrubOwned` (owned, for a decode holding a live `&Map` into its tree). This replaces the three near-identical guards the crate had grown in `codec::fields` and `seal::body`, so the terminal-owner rationale is stated once at the mechanism instead of at each copy.
- **`codec::decode`** — the array accumulator, the map accumulator, and the top-level value all stand under `ScrubOnDrop`. The map arm keeps its single `Vec<(String, Value)>`: the guard wipes the values and leaves the keys, which are field names and never secret.
- **`seal::body::PreservedFields`** — a `Map`-backed carrier replacing `Vec<(String, Value)>` across the seal layer. It wipes on `Drop`, renders redacted, and makes duplicate preserved keys unrepresentable. Because `collect_unknown`/`merge_unknown` are shared, one carrier covers `ChildRef`, `Version`, `ReadBody`, `Envelope`, and the grant/section/write-body structs at once.
- **`Map::zeroize_bytes` marks the map**, and `Map::reject_if_wiped` — release-active, returning `Malformed::WipedMap` — runs on both encoder passes and on `encode_map_partial`'s merged head, so the reject fires at every nesting level before a byte is allocated.
- **`seal::envelope::set_grant_section`** — a writer to match the existing `grant_section_bytes`/`has_grant_section` readers, so the `grantSection` key and its byte-string shape live in core rather than being spelled at each engine call site. `PreservedFields::insert` is `pub(crate)`, which narrows what was previously a fully public `Vec` field.
- **`encode_read_body`** re-checks child uniqueness release-active rather than under a `debug_assert!` a release build strips (AGENTS.md rule 8). Not live — `seal_read_body` already calls `validate()?` first — but `encode_read_body` is public.

No wire-format change; the KAT tree is byte-identical.

### Note for #937

#937 (redact `Value`, `Map`, and `Version` renderings) is not started and touches these types, so this PR deliberately stays off that surface: `Value`'s `Display`, `Map`'s derived `Debug`, and `Version`'s hand-written `Debug` are untouched. The only rendering this PR adds is `PreservedFields`' own redacted `Debug`, which shares a new `fmt_redacted_keys` helper with `codec::UnknownFields`. #937 should rebase cleanly.

One thing the crypto gate found that #937 as scoped does **not** reach: `ChildRef` still derives `Debug`, so a `{:?}` on a `ReadBody::Folder` renders every plaintext filename and `ipnsName`. Pre-existing and out of scope here — raised on #937 to widen it.

## Tests

Each new test is confirmed red when its fix is reverted, verified by stubbing out `Map::zeroize_bytes`'s mark, `PreservedFields::Drop`, `ScrubOnDrop::drop`, and `encode_read_body`'s `validate()?`, then re-running `cargo test -p cipherbox-core`:

| Test | Fires when |
| ---- | ---------- |
| `codec::scrub::tests::a_borrowed_guard_wipes_what_it_guards` | the guard stops wiping the accumulator a reject abandons |
| `codec::encode::tests::a_wiped_map_refuses_to_encode` | a wiped map nested inside a body encodes instead of rejecting |
| `codec::fields::tests::merged_encode_rejects_a_wiped_known_side` | the partial encoder emits from a scrubbed known side |
| `seal::body::tests::preserved_fields_own_the_wipe_of_what_a_decode_cloned_out` | the carrier stops scrubbing the clones it holds |
| `seal::body::tests::encode_rejects_a_body_decode_would_refuse` | the encode-side uniqueness check goes back to being release-stripped |
| `codec_props::the_public_encoder_refuses_a_wiped_map` | the `wiped-map` verdict stops being reachable across the crate boundary |

Also pinned: `seal::body::tests::preserved_fields_debug_redacts_its_values`, and `kat_manifest::every_crate_check_is_pinned_by_a_vector_family` now expects two unit-pinned encoder caller-bug checks rather than one.

The `Drop` wiring on `PreservedFields` is not observable in safe Rust — same as `UnknownFields`' `Zeroizing`, which #902 also left unpinned. The tests pin the wipe it calls.

## Verification

All exit 0:

- `cargo fmt --all --check`
- `cargo test --workspace` — 32 suites green
- `cargo test --release -p cipherbox-core` — the fail-closed checks are release-active
- `pnpm -r --if-present run typecheck`
- `pnpm -r --if-present run test` — 111 client, 177 api, 67 web
- `eslint .`

Review gates run on `git diff main...HEAD`: `/security-review` found nothing at confidence 8 or above. `/simplify` ran four parallel passes — the guard unification, the `Map`-backed carrier, the shared redaction helper, the reverted accumulator split, and the doc trimming above are its findings folded in. `/crypto-privacy-review` rated the diff LOW risk with no CRITICAL or HIGH findings; the `encode_fixed_depth` panic doc, the `Map` mark-boundary note, the release-active `encode_read_body` check, and the public-API `wiped-map` pin are its findings folded in.

Deferred with a dependency edge: #956 — extend the scrub guard to the write-body, section, and mailbox codecs, which strand pseudonymous identifiers and already-sealed bytes rather than key material.

Closes #936
Part of #655

## Summary by CodeRabbit

* **New Features**
  * Added a preserved-fields type for safely retaining unknown data during encoding and decoding.
  * Added a supported way to set grant-section data on envelopes.

* **Bug Fixes**
  * Prevented encoding of data that has already been securely wiped.
  * Improved cleanup of partially decoded values after decoding failures.
  * Added validation to reject invalid or duplicate fields consistently, including in release builds.

* **Security**
  * Enhanced zeroization of sensitive buffers and transient decoded data.
  * Redacted preserved-field keys in debug output.

### DIFF
--- a/crates/core/examples/kat_gen.rs
+++ b/crates/core/examples/kat_gen.rs
@@ -35,10 +35,10 @@ use cipherbox_core::seal::{
     self, AAD_DOMAIN, AadContext, AscentLink, CONTENT_KEY_HPKE_INFO, CONTENT_KEY_V, ChildRef,
     ChildScopeRef, GrantBlobPayload, GrantLedgerEntry, GrantSetCommitment, GrantSetEntry,
     HistoryLinkPayload, NodeKind, OP_RECORD_HPKE_INFO, OP_RECORD_V, OpRecordHeader,
-    OverrideSeedPayload, OwnerWriteBlobPayload, Permission, ReadBody, SETTINGS_RECORD_HPKE_INFO,
-    SETTINGS_RECORD_V, STRUCT_TAG_ASCENT_LINK, STRUCT_TAG_CONTENT_KEY, STRUCT_TAG_GRANT_BLOB,
-    STRUCT_TAG_HISTORY_LINK, STRUCT_TAG_MAILBOX_PAYLOAD, STRUCT_TAG_OP_RECORD,
-    STRUCT_TAG_OWNER_BLOB, STRUCT_TAG_OWNER_WRITE_BLOB, STRUCT_TAG_READ_BODY,
+    OverrideSeedPayload, OwnerWriteBlobPayload, Permission, PreservedFields, ReadBody,
+    SETTINGS_RECORD_HPKE_INFO, SETTINGS_RECORD_V, STRUCT_TAG_ASCENT_LINK, STRUCT_TAG_CONTENT_KEY,
+    STRUCT_TAG_GRANT_BLOB, STRUCT_TAG_HISTORY_LINK, STRUCT_TAG_MAILBOX_PAYLOAD,
+    STRUCT_TAG_OP_RECORD, STRUCT_TAG_OWNER_BLOB, STRUCT_TAG_OWNER_WRITE_BLOB, STRUCT_TAG_READ_BODY,
     STRUCT_TAG_SETTINGS_RECORD, STRUCT_TAG_WRITE_BODY, STRUCT_TAGS, SettingsRecordHeader,
     SignedAscentLink, SignedGrantBlob, SignedOwnerBlob, SignedSealed, StructureSigInput, Version,
     WriteBody, build_aad, content_key_aad, decode_ascent_link, decode_envelope,
@@ -2741,7 +2741,7 @@ fn sample_folder() -> ReadBody {
                 ipns_name: b"ipns-name-a".to_vec(),
                 kind: NodeKind::File,
                 link_counter: 1,
-                unknown: Vec::new(),
+                unknown: PreservedFields::new(),
             },
             ChildRef {
                 id: [0x22; 16],
@@ -2749,10 +2749,10 @@ fn sample_folder() -> ReadBody {
                 ipns_name: b"ipns-name-b".to_vec(),
                 kind: NodeKind::Folder,
                 link_counter: 2,
-                unknown: Vec::new(),
+                unknown: PreservedFields::new(),
             },
         ],
-        unknown: Vec::new(),
+        unknown: PreservedFields::new(),
     }
 }
 
@@ -2765,7 +2765,7 @@ fn sample_file() -> ReadBody {
             Version::new(b"content-cid-new".to_vec(), [0x77; 32], 8192, 2500),
             Version::new(b"content-cid-old".to_vec(), [0x66; 32], 4096, 1500),
         ],
-        unknown: Vec::new(),
+        unknown: PreservedFields::new(),
     }
 }
 
@@ -2940,13 +2940,13 @@ fn build_read_body_accept() -> Vec<ReadBodyAcceptVector> {
         created_at: 1,
         modified_at: 2,
         children: Vec::new(),
-        unknown: Vec::new(),
+        unknown: PreservedFields::new(),
     };
     let single_version = ReadBody::File {
         created_at: 3,
         modified_at: 4,
         versions: vec![Version::new(b"cid".to_vec(), [0x55; 32], 512, 4)],
-        unknown: Vec::new(),
+        unknown: PreservedFields::new(),
     };
     let cases: Vec<(&str, ReadBody)> = vec![
         ("empty-folder", empty_folder),
@@ -4526,7 +4526,7 @@ fn build_write_body_accept() -> Vec<WriteBodyAcceptVector> {
             ChildScopeRef::new([0x55; 16], b"child-scope-ipns-a".to_vec()),
             ChildScopeRef::new([0x66; 16], b"child-scope-ipns-b".to_vec()),
         ],
-        unknown: Vec::new(),
+        unknown: PreservedFields::new(),
     };
     // Write epoch 1: no prior write epoch, so no history link, and a scope root
     // with no descendant scopes and no grants yet.
@@ -4534,7 +4534,7 @@ fn build_write_body_accept() -> Vec<WriteBodyAcceptVector> {
         grant_ledger: Vec::new(),
         write_history_link: Vec::new(),
         direct_child_scope_index: Vec::new(),
-        unknown: Vec::new(),
+        unknown: PreservedFields::new(),
     };
     let read_only = WriteBody {
         grant_ledger: vec![GrantLedgerEntry::new(
@@ -4545,7 +4545,7 @@ fn build_write_body_accept() -> Vec<WriteBodyAcceptVector> {
         )],
         write_history_link: b"h".to_vec(),
         direct_child_scope_index: vec![ChildScopeRef::new([0x77; 16], b"one-child".to_vec())],
-        unknown: Vec::new(),
+        unknown: PreservedFields::new(),
     };
     let cases: Vec<(&str, WriteBody)> = vec![
         ("full", full),
@@ -4674,7 +4674,7 @@ fn section_commitment(entries: Vec<GrantSetEntry>) -> GrantSetCommitment {
         ipns_name: b"grant-section-scope-root".to_vec(),
         owner_pseudonym_pk: [0x88; 32],
         entries,
-        unknown: Vec::new(),
+        unknown: PreservedFields::new(),
     }
 }
 
@@ -4684,7 +4684,7 @@ fn section_grant_blob(tag: u8) -> SignedGrantBlob {
         enc: [0x0a; 32],
         ciphertext: vec![0x0b, 0x0c, 0x0d],
         signature: [0x0e; 64],
-        unknown: Vec::new(),
+        unknown: PreservedFields::new(),
     }
 }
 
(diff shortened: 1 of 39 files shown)


FSM1:
## CodeRabbit review 4840313926 — resolved

One nitpick, body-only, no review threads. Swept `reviewThreads` and `greptile-apps`: none open.

### `crates/core/src/codec/scrub.rs:99-109` — valid, fixed differently than suggested

**The finding is correct.** `an_owning_guard_wipes_the_tree_it_took` asserted only pre-drop readability and nothing about the wipe. Confirmed empirically: with `ScrubOwned::drop` gutted to `fn drop(&mut self) {}`, the original test body still passed.

**The proposed patch does not fix it.** It calls `guard.0.zeroize_bytes()` by hand and asserts the result — that exercises `zeroize_bytes`, not `Drop`. Deleting the call from the `Drop` impl leaves that test green, so the exact regression it names still slips through.

The real difficulty: `ScrubOwned` *owns* its value, so after drop nothing is left to read soundly. The sibling `ScrubOnDrop` can assert post-drop only because it holds `&mut` and the data outlives the guard. Reading the freed buffer through a pointer captured pre-drop is UB, so it is not an option here.

**What landed instead** — `ScrubOwned` now routes through the same `Scrub` trait its borrowed sibling already uses: `ScrubOwned<T: Scrub>`, with `drop` calling `self.0.scrub()` instead of `Value::zeroize_bytes` directly. Every production site is `ScrubOwned(decode(bytes)?)` and infers `ScrubOwned<Value>`, so no call site changed, and the two guards stop disagreeing about what a guard wipes.

That makes the wipe observable from the scrubbed value's own side: `an_owning_guard_wipes_the_tree_it_took` now guards a `SpyTree` whose `scrub()` sets a borrowed `Cell`, asserting it flips only once the guard falls. The old body is kept and honestly renamed to `an_owning_guard_lends_out_the_tree_it_took` — it covers `value()`, not `Drop`.

Coverage chain: *Drop invokes `scrub`* (new test) composed with *`<Value as Scrub>::scrub` zeroizes nested buffers* (`a_borrowed_guard_wipes_what_it_guards`, unchanged).

**Red-before-green, as observed.** With `Drop` emptied:

```
test codec::scrub::tests::an_owning_guard_wipes_the_tree_it_took ... FAILED
panicked at crates/core/src/codec/scrub.rs:126:9:
the tree the guard owns is wiped on drop
test result: FAILED. 2 passed; 1 failed
```

Call restored: `3 passed; 0 failed`.

**On the allocator-watchdog alternative** (`crates/engine/tests/content_wipe.rs`): considered and rejected here. That watchdog owns its whole test binary, so it can only live in an integration test — and `ScrubOwned` is `pub(crate)`, unreachable from `crates/core/tests/`. A core-side watchdog would have to drive a public seal API and would assert the whole decode pipeline strands no plaintext, which is a different (and already separately covered) property, not this unit's `Drop`. It earns its `unsafe` in `content_wipe.rs` because stranded plaintext across a pipeline has no unit-level expression; "Drop calls scrub" does.

**Residual, stated plainly:** these tests prove `Drop` scrubs and that scrubbing zeroizes the tree's buffers in place. They do not independently re-prove that the heap block `Value::Bytes`'s `Vec` frees is zero — that is `Vec::zeroize`'s contract (zeroes the full capacity in place, keeping the allocation), not ours.

Gates: `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo test -p cipherbox-core` 218 + 8 + 2 + 69 + 2 + 5 passed, 0 failed; `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown` clean.

PR #968, opened 2026-08-02
fix: harden the desktop floor store read-modify-write and intent length prefix

### LINKED ISSUE #704: desktop-seams: floor_store RMW single-writer hardening + 32-bit intent overflow
From the retro `/security-review` of #690 (merged). **Severity: LOW (defense-in-depth; neither reachable today).** File: `crates/desktop-seams/src/floor_store.rs`.

1. **Unserialized read-modify-write in `raise_floor`** (L81-91, 165-176). Two *concurrent* `commit_floors` raising the same epoch key could interleave read->read->write(hi)->write(lo) and durably regress a floor with no intent left to heal it. Not reachable today (single-threaded `spawn_local` LocalSet, single-writer engine, no `.await` between floor read and write), but the new unique-intent-filename mechanism explicitly defends against overlapping commits — inconsistent with leaving the RMW unguarded. Recommend a `debug_assert!` single-writer marker or an internal `Mutex` around the RMW so the two hardening decisions are consistent.

2. **32-bit intent overflow** (L203-204). `intent_slice` computes `start + len` on a corrupt-controlled `key_len`, which can overflow `usize` on a 32-bit target and panic (debug) instead of returning the clean `corrupt_intent()` SeamError. Not reachable on 64-bit desktop targets; even the panic is fail-closed. Tidy: `start.checked_add(len).and_then(|end| bytes.get(start..end))`.

Part of #655

FSM1:
Closes #704

Retro `/security-review` hardening on `crates/desktop-seams/src/floor_store.rs`. Both items are LOW / defence-in-depth. I re-verified the reachability claims rather than taking them on trust — both hold — so this makes the store's monotonic-floor invariant structural rather than incidental, without changing observable behaviour.

## 1 — Unserialized read-modify-write in `raise_floor`

`raise_floor` read the stored floor, took `max`, then wrote, with nothing serializing the two halves. Two concurrent raises on one key could interleave read/read/write-high/write-low and durably regress the floor. That loss is unhealable: both commits return `Ok` and clear their own intent records, so replay has nothing left to roll forward.

Fix: a `Mutex<()>` on `FileFloorStore`, held across the whole RMW in `raise_floor` — `floor_store.rs:38-42` and `:88-92`. A poisoned lock fails closed with a `SeamError`.

Per-key serialization rather than per-batch is deliberate and sufficient. Monotonic-max is a per-key invariant, and the seam contract makes the desktop store roll-forward rather than transactional across keys, so an interleave between the keys of two batches heals idempotently on replay. `apply_raises` and `replay_intents` inherit the lock through `raise_floor`; there is one acquisition site, so no nesting and no deadlock edge. The guard never spans an `.await` — `raise_floor` is a sync fn.

Scope boundary: the lock is in-process, per handle. Two `FileFloorStore` handles over the same directory, or two processes, are not serialized by it. The engine constructs one store per session, and cross-process file locking is well beyond this issue.

## 2 — 32-bit intent overflow in `intent_slice`

`intent_slice` computed `start + len` on a `len` taken from the corrupt-controlled 4-byte length prefix, which can wrap `usize` on a 32-bit target and panic in debug rather than returning the clean `corrupt_intent()` error. Fix: `start.checked_add(len).and_then(|end| bytes.get(start..end))` — `floor_store.rs:223-231`.

While there, the encode side had the mirroring gap that AGENTS.md rule 8 requires closing: `encode_intent` wrote `raise.key.len() as u32`, a silently truncating cast. A truncated prefix does not merely produce a record the decoder rejects — it reframes the record, so the decoder reads a short key and then parses key bytes as the next raise's value. `encode_intent` now returns `SeamResult<Vec<u8>>` and rejects through `intent_key_len` — `floor_store.rs:190-197`. Release-active `Err`, no `debug_assert!`.

## Reachability — verified, neither is live

- Item 1: `Scheduler` spawns via `tokio::task::spawn_local` (`scheduler.rs:19`, `:49`), which requires a `LocalSet` on a current-thread runtime, so engine tasks are single-threaded. `raise_floor` is a sync fn with no `.await` between read and write, so a task cannot yield mid-RMW even under a multi-threaded executor. `FileFloorStore` is also not yet wired into a shell — the only in-tree constructions are the `lib.rs` re-export and the conformance suite.
- Item 2: 32-bit is not a supported desktop target, and even the panic is fail-closed.
- The encode-side truncation needs a 4 GiB key; keys are scope ids and `ipnsName` bytes.

## Tests

- `concurrent_raises_never_regress_a_floor` — 32 keys, two barrier-synced threads per key raising 2 and 1. Negative control: with the mutex removed this fails immediately with a durable `Some(1)` where `Some(2)` is required, reproducing the exact regression the issue describes. It passes deterministically with the fix.
- `intent_slice_fails_closed_on_an_overflowing_end` — the `usize` boundary, on any pointer width.
- `decode_rejects_an_oversized_key_length` — a record claiming `u32::MAX` key bytes.
- `encode_rejects_a_key_longer_than_its_length_prefix` — the `u32::MAX` encode boundary. `intent_key_len` takes a `u64` precisely so this is testable without a 4 GiB allocation, and so the test is pointer-width independent.

All four are ordinary `#[test]`s, so they fire in a release build.

## Gates

All exit 0, rebased onto `main` at 20d03b873, with a per-worktree `CARGO_TARGET_DIR`:

- `cargo fmt --all --check`
- `cargo test -p cipherbox-engine -p cipherbox-core -p cipherbox-desktop-seams` — all green; `cipherbox-desktop-seams` 19 lib + 13 conformance passed
- `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown`

## Parallel work

Confined to `floor_store.rs`. It does not touch `fs_util.rs`, `staging_store.rs`, or `tests/conformance.rs`, so it does not collide with #952.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented concurrent floor updates from overwriting newer persisted values.
  * Added validation for malformed or oversized intent data.
  * Improved handling of overflow conditions by failing safely instead of attempting invalid reads.

* **Tests**
  * Added coverage for concurrent updates, oversized lengths, and arithmetic overflow scenarios.

### DIFF
--- a/crates/desktop-seams/src/floor_store.rs
+++ b/crates/desktop-seams/src/floor_store.rs
@@ -35,10 +35,9 @@ pub struct FileFloorStore {
     epoch_dir: PathBuf,
     seq_dir: PathBuf,
     intent_dir: PathBuf,
-    /// Serializes the read-modify-write in [`Self::raise_floor`]. Concurrent
-    /// raises on one key could otherwise interleave read/read/write-high/
-    /// write-low and durably regress the floor — a loss no intent record can
-    /// heal, because both commits succeeded and cleared their own records.
+    /// Serializes the read-modify-write in [`Self::raise_floor`] within this
+    /// handle: interleaved raises can durably regress a floor, and intent
+    /// replay cannot heal that loss (#704).
     write_lock: Mutex<()>,
 }
 
@@ -187,9 +186,8 @@ impl FloorStore for FileFloorStore {
     }
 }
 
-/// The 4-byte little-endian key-length prefix, rejecting a key the prefix
-/// cannot represent. Truncating instead would reframe the record: the decoder
-/// would read a short key and then parse key bytes as the next raise's value.
+/// The 4-byte little-endian key-length prefix. Fails closed on a key the
+/// prefix cannot represent: a truncated length reframes the whole record.
 fn intent_key_len(len: u64) -> SeamResult<[u8; 4]> {
     u32::try_from(len)
         .map(u32::to_le_bytes)
@@ -220,8 +218,8 @@ fn corrupt_intent() -> SeamError {
 }
 
 /// Reads a fixed slice at `start`, failing closed if it runs past the record.
-/// The end offset is checked: `len` comes from the corrupt-controlled length
-/// prefix, so on a 32-bit target it can wrap `usize`.
+/// `len` is corrupt-controlled, so `start + len` can wrap `usize` on a 32-bit
+/// target.
 fn intent_slice(bytes: &[u8], start: usize, len: usize) -> SeamResult<&[u8]> {
     start
         .checked_add(len)
@@ -276,7 +274,6 @@ mod tests {
         );
     }
 
-    /// A record claiming more key bytes than it carries fails closed.
     #[test]
     fn decode_rejects_an_oversized_key_length() {
         let mut bytes = vec![INTENT_TAG_EPOCH];
@@ -284,28 +281,19 @@ mod tests {
         assert!(decode_intent(&bytes).is_err());
     }
 
-    /// An end offset past `usize` fails closed rather than wrapping into an
-    /// in-range slice — the length is corrupt-controlled, and on a 32-bit
-    /// target it can exceed what the start offset leaves.
     #[test]
     fn intent_slice_fails_closed_on_an_overflowing_end() {
-        assert!(intent_slice(b"abc", usize::MAX, 1).is_err());
         assert!(intent_slice(b"abc", 1, usize::MAX).is_err());
     }
 
-    /// Encode fails closed on a key the 4-byte prefix cannot represent instead
-    /// of truncating it — a truncated prefix reframes the record and the
-    /// decoder would parse key bytes as the next raise.
     #[test]
     fn encode_rejects_a_key_longer_than_its_length_prefix() {
         assert!(intent_key_len(u32::MAX as u64).is_ok());
         assert!(intent_key_len(u32::MAX as u64 + 1).is_err());
     }
 
-    /// Two writers racing on one key cannot interleave read/read/write-high/
-    /// write-low: whichever lands second re-reads the raised floor and keeps
-    /// it. A durable regression here is unhealable — both commits return `Ok`
-    /// and clear their own intent records, leaving nothing for replay.
+    /// The #704 regression: with the raises unserialized, the low one can read
+    /// the pre-raise floor and land last.
     #[test]
     fn concurrent_raises_never_regress_a_floor() {
         let dir = tempfile::tempdir().unwrap();
--- a/crates/desktop-seams/src/floor_store.rs
+++ b/crates/desktop-seams/src/floor_store.rs
@@ -2,6 +2,7 @@
 //! write-ahead intent record for cross-key atomic batch commits.
 
 use std::path::{Path, PathBuf};
+use std::sync::Mutex;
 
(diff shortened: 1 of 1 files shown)


FSM1:
## Review sweep — rebased onto `main`, one body-only finding, declined with measurements

Rebased onto `main` at 20d03b873 (8 commits). No textual conflict, and no semantic one either: the diff is confined to `crates/desktop-seams/src/floor_store.rs`, which `main` did not touch in that range — #952 landed in `desktop-seams` but in `fs_util.rs`/`staging_store.rs`. The branch constructs no preserved-unknowns field and implements no seam trait whose members moved, so neither the `PreservedFields` change (#957) nor the `HttpSeam` surface reaches it.

Gates after the rebase, all exit 0:

- `cargo fmt --all --check`
- `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown`
- `cargo test -p cipherbox-engine -p cipherbox-core -p cipherbox-desktop-seams` — 1133 passed, 0 failed (`cipherbox-desktop-seams` 19 lib + 13 conformance)

Sweep covered `reviewThreads` (zero inline threads), the submitted review bodies, and the issue comments. One reviewer (CodeRabbit); no Greptile review. One finding, in the review body only.

### Declined — "Make the concurrency regression deterministic" — `floor_store.rs:298-322`, CodeRabbit nitpick, body-only

The claim is that `concurrent_raises_never_regress_a_floor` "can pass without the lock" because the barrier only synchronizes the *start* of the two raises, and the fix is a test-only hook after `read_floor` to pause the low raise.

I measured it rather than reasoning about it. Removing only the `write_lock` acquisition from `raise_floor` and running the test 40 times:

- **lock removed: 0 passed, 40 failed** — every run trips `left: Some(1), right: Some(2)` at `floor_store.rs:313`
- **lock present: 40 passed, 0 failed**

The test discriminates on the lock perfectly, so the premise does not hold in practice. The reason it is not a coin flip is asymmetry in the two halves of the race: the barrier releases both threads into `read_floor`, a single small open-read-close, while the half that has to win for the test to pass falsely is `atomic_write` — temp file, write, `sync_all`, rename, parent-dir fsync. The read/read interleave is not a narrow window, it is the overwhelmingly likely schedule, and the assertion only needs 1 of the 32 independent keys to lose.

The more important property: **this test cannot go flaky-red.** With the lock in place the assertion is guaranteed by the mutex under every schedule, so no CI run can fail on it. The only thing at stake is its strength as a regression detector, and that is measured above at 40/40.

Against that, the proposed fix costs a `#[cfg(test)]` callback field on `FileFloorStore` and a branch inside the production `raise_floor` whose sole consumer is one test — the speculative-machinery-in-production shape AGENTS.md and `/simplify` reject. It would also weaken what the test proves: a forced pause asserts that the hook fires where it was planted, not that the critical section spans the read and the write. CodeRabbit tags the item Trivial / Heavy lift itself.

No code change. Happy to revisit if the test is ever observed green on a build without the lock.

### Also considered and left alone

`apply_raises` takes the mutex once per key rather than once per batch, so a batch is N acquisitions and two concurrent `commit_floors` can interleave between keys. Hoisting the lock to the batch would need an inner non-locking `raise_floor` to avoid self-deadlock on the non-reentrant `std::sync::Mutex`, and it buys no correctness: floors are monotonic-max per key, so any interleaving of two batches converges on the same per-key maxima, a concurrent reader can only observe an earlier valid floor, and crash recovery is roll-forward. The acquisitions are uncontended nanoseconds against a millisecond fsync, so it is not the cost centre either. The extra indirection would be the change `/simplify` pushes back on.

Issue #975, opened 2026-08-02
Nothing typechecks packages/client/test/browser so drifted test doubles pass silently


FSM1:
## Defect

No tsconfig in the repo covers `packages/client/test/browser/`, so nothing typechecks the Playwright browser-suite harness.

Surfaced while consolidating the `EngineHostLike` test doubles during the review-gate pass on #641 (PR #951): the media worker's double was **missing `siweChallenge` entirely** and no gate caught it. A test double that does not implement its interface compiles and runs, and the suite passes, because `tsc` never looks at the directory.

## Why it matters

`test/browser/*` is not incidental scaffolding — it hosts the engine-worker doubles, the seam conformance workers and the leadership/media harnesses that the `Client Browser Suite` CI gate runs. A drifted double there produces a suite that exercises the wrong shape while reporting green, which is the same failure class as a silently-skipped test file.

The repo has been bitten by the adjacent version of this before: `apps/web`'s vitest `include` matches `*.test.ts` only, so `*.spec.ts` files were silently skipped in CI.

## Scope

- Add `test/browser` to a tsconfig that the `Typecheck` CI job actually runs — either extend `packages/client/tsconfig.json`'s `include`, or add a dedicated `tsconfig.browser-test.json` referenced from the typecheck script.
- Fix whatever the first clean run surfaces. Expect drift beyond the one already found.
- Confirm the new coverage is load-bearing: deliberately break a double's signature and check the typecheck gate goes red.

Note `packages/client/src/testkit.ts` is build-excluded but still typechecked; `test/browser` should end up in the same position.

Part of #655

PR #978, opened 2026-08-02
test(client): cover the media range window arithmetic and seed each tab distinctly

### LINKED ISSUE #963: web-client: unit-cover the media range arithmetic and give each tab distinct fixture bytes
## Problem

Two test-coverage residuals from the byte-pipe slice, both in the media path.

### `resolveMediaRequest` arithmetic is unit-covered nowhere

`packages/client/src/media/range.ts` resolves every media request the Service Worker pipe serves: it parses the `Range` header, decides 200/206/416, and computes both the `content-range` header and the plaintext `window` the broker then streams. `range.test.ts` asserts only the content-type downgrade and the hardening headers — it never reads `head.window` and never asserts a `content-range` value.

So the arithmetic deciding which bytes a viewer actually receives has no unit coverage. Branches currently unasserted:

- **suffix form** `bytes=-N` — clamps to `Math.max(0, totalSize - wanted)`, so a suffix larger than the file yields the whole file, not a 416.
- **suffix degenerate cases** — `bytes=-0` and a zero-length file are both 416.
- **open interval** `bytes=N-` — `last` becomes `totalSize - 1`.
- **`last` past EOF** `bytes=N-M` with `M >= totalSize` — clamped by `Math.min(requestedLast, totalSize - 1)`.
- **416 paths** — `offset >= totalSize`, and an inverted interval such as `bytes=100-50`.
- **whole-file fall-through** — multi-range (a comma in the set), a non-`bytes` unit, and a malformed spec all return 200 with the whole file rather than a 416, which is deliberate and untested.
- **unsafe-integer fail-closed** — `parseOffset` returns `null` for a digit run past `Number.MAX_SAFE_INTEGER`, which must be a 416, not a clamp.

The browser suite covers the 200/206/416 outcomes end to end, so this is under-unit-tested rather than unverified — but the per-branch window arithmetic is exactly what a unit test is for, and a regression there is silent data corruption in playback.

### Both browser-suite tabs synthesize identical bytes

`packages/client/test/browser/media.ts` hardcodes `TAB_SEED` in its `downloadRange` stub (line 71), so every tab built by `localTab()` in `media.spec.ts` produces the same fixture bytes. The two-tab test asserts both bodies against `fixtureSlice(..., TAB_SEED)`, so a cross-tab ticket or registry mix-up that served tab B's stream out of tab A's registry would still pass every byte assertion — only the 404 on a foreign ticket catches it today.

`mediaFixture.ts` already exports `LEADER_SEED` beside `TAB_SEED`, so the seam exists. `localTab` needs to accept a seed and thread it into the stub; the callers that do not care keep the current default.

## Gate

The `packages/client` vitest suite: `range.test.ts` asserts `window` and `content-range` across the branch list above, including the unsafe-integer 416. The `packages/client` browser suite: the two-tab test drives distinct seeds per tab and asserts each body against its own tab's fixture.

## Depends on

#641 — `range.ts`, `media.ts`, and the two-tab spec all land there.

Part of #655

FSM1:
Closes #963

Two test-coverage residuals from the byte-pipe slice, both in the media path.

## `resolveMediaRequest` window arithmetic

`range.test.ts` asserted only the content-type downgrade and the hardening headers — it never read `head.window` and never asserted a `content-range` value, so the arithmetic deciding which plaintext bytes a viewer receives had no unit coverage. Three new table-driven blocks close that:

- **206 windows** — plain interval, last-byte boundary, open interval, a `last` past EOF clamped to `totalSize - 1`, the suffix form, a suffix longer than the file resolving to the whole file rather than a 416, plus case-insensitive and whitespace-padded specs. Each row asserts the exact `window`, the `content-range`, and a `content-length` that agrees with the window.
- **Whole-file fall-through** — absent, blank, multi-range, non-`bytes`, and malformed specs all answer 200 with the whole file and no `content-range`; a zero-byte file yields an empty window.
- **416** — offset at EOF, inverted interval, `bytes=-0`, any range on an empty file, and all three digit runs past `Number.MAX_SAFE_INTEGER` that `parseOffset` fails closed on. Every 416 row asserts `'window' in head === false`: a rejected range must read no plaintext at all.

Verified by mutation: thirteen single-line mutations of `range.ts` — dropping the `Math.min` EOF clamp, dropping the `Math.max(0, …)` suffix clamp, each of the three `parseOffset` null gates, the inverted-interval gate, the open-interval `totalSize - 1`, the `last - offset + 1` length, the multi-range fall-through, the `wanted === 0` and `totalSize === 0` suffix gates, the set `trim()`, and the `toLowerCase()` — each turn at least one new test red.

## Distinct fixture bytes per tab

`media.ts` hardcoded `TAB_SEED` in its `downloadRange` stub, so both tabs in `two open tabs each stream from their own broker` synthesized identical bytes: a cross-tab stream mix-up that answered one tab's ticket out of the other tab's broker would still have passed every byte assertion. `MediaStartOptions` now carries a `seed` and `localTab` threads it, so tab A serves `TAB_SEED` and tab B `SECOND_TAB_SEED`. Mutation-checked the same way — pinning the harness back to `TAB_SEED` turns the two-tab test red.

## Gates

- `pnpm typecheck` — pass, all four workspace projects.
- `pnpm --filter @cipherbox/client test` — 22 files, 252 tests, pass. Gated on merge by the root `pnpm test` fan-out, so no workflow change is needed.
- `pnpm --filter @cipherbox/client run test:browser` — 27 passed, including all 8 media specs.

`packages/client/test/browser` is not typechecked today (#975), so the harness changes there were verified by running them.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved media range handling, including partial content responses, empty files, whole-file requests, and invalid range rejection.
  * Strengthened media response validation and security-header coverage.
  * Improved multi-tab media isolation by ensuring each session serves distinct fixture data.

### DIFF
--- a/packages/client/src/media/range.test.ts
+++ b/packages/client/src/media/range.test.ts
@@ -1,10 +1,19 @@
 import { describe, expect, it } from 'vitest';
 
-import { resolveMediaRequest, type MediaHead } from './range.js';
+import { resolveMediaRequest, type MediaHead, type MediaWindow } from './range.js';
 
 const SIZE = 4096;
+const MIME = 'video/mp4';
+
+/** The first integer past `Number.MAX_SAFE_INTEGER`. */
+const UNSAFE = '9007199254740992';
 
 const headerMap = (head: MediaHead): Map<string, string> => new Map(head.headers);
+
+const windowOf = (head: MediaHead): MediaWindow => {
+  if (head.status === 416) throw new Error('a 416 head carries no window');
+  return head.window;
+};
 
 const contentTypeFor = (mimeType: string): string | undefined =>
   headerMap(resolveMediaRequest(null, SIZE, mimeType)).get('content-type');
@@ -31,9 +40,9 @@ describe('resolveMediaRequest content-type', () => {
 
 describe('resolveMediaRequest hardening headers', () => {
   const cases: Array<[string, MediaHead]> = [
-    ['200', resolveMediaRequest(null, SIZE, 'video/mp4')],
-    ['206', resolveMediaRequest('bytes=0-99', SIZE, 'video/mp4')],
-    ['416', resolveMediaRequest(`bytes=${SIZE}-`, SIZE, 'video/mp4')],
+    ['200', resolveMediaRequest(null, SIZE, MIME)],
+    ['206', resolveMediaRequest('bytes=0-99', SIZE, MIME)],
+    ['416', resolveMediaRequest(`bytes=${SIZE}-`, SIZE, MIME)],
   ];
 
   for (const [status, head] of cases) {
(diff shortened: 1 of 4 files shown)


FSM1:
## Review triage — CodeRabbit review of 2026-08-03

One finding, body-only, no inline threads. Verdict: **declined**, with evidence.

### Finding

> `packages/client/test/browser/media.spec.ts` `99-100` — **Remove or replace this descriptive helper comment.** The comment states what `localTab` does. The helper name and implementation already show that behavior. Remove it, or state a non-obvious reason for the helper. As per coding guidelines, "Keep comments short and explain why rather than what; avoid comments ... narrating obvious behavior."

### Declined — reasoning

1. **The quoted guideline text does not exist in this repo.** The nearest real rule is `AGENTS.md` §Code Style, which targets paragraph-length apologia, absence-justifying comments, and unchanged-code apologia, plus length discipline aimed at multi-line prose. A one-line JSDoc summary on a test-harness helper is not the shape that rule is aimed at.

2. **The line is pre-existing.** Before this PR it read `/** A controlled tab whose broker serves the tab-seeded fixture. */`. This diff edits it only to stay accurate after `localTab` gained a `seed` parameter. Deleting it is unrelated cleanup, not a defect this change introduces.

3. **It records a cross-file binding the four-line body does not show.** `seed` is threaded `localTab` to `tab.start('local', seed)` to `cbMediaStart` to `localReader(seed ?? TAB_SEED)` in `test/browser/media.ts` to `fixtureBuffer(offset, length, seed)`. That this is the same seed the spec's `fixtureSlice(0, size, SEED)` assertions consume resolves two files away, including the `TAB_SEED` default that the seedless call sites rely on.

4. **The alternative the nitpick offers would violate a different rule.** The genuinely non-obvious rationale behind `localTab` — a registration does not control the tab until it claims, so a fetch before `awaitControl()` misses the pipe — already lives at its home, `test/browser/media.ts:111`. `AGENTS.md`: "State genuine non-obvious domain rationale **once**, at its home ... not restated on every caller." Restating it here is exactly the caller-side duplication that rule forbids.

5. **Consistency.** The browser-suite harness family carries 17 of these one-line docs, 7 of them in `test/browser/media.ts` and 3 in `test/browser/mediaFixture.ts` — both files this same PR edits. Removing this one leaves an inconsistent island.

6. **No behavioral surface.** It is a comment; accepting or declining cannot change what the suite proves.

No code change taken from this review.

### Rebase and re-verification

Rebased onto `main` at `4044adbcd`, clean, no conflicts. New tip `75d3ca34d`. Nothing in this branch implements `HttpSeam`, so the `sendCapped` interface addition from #951 does not bite; no Rust in scope.

- `pnpm -r --if-present run typecheck` — exit 0, all four projects Done.
- `pnpm -r --if-present run test` — exit 0: client 265, api 178, web 69.
- `pnpm --filter @cipherbox/client run test:browser` — exit 0, **28 passed**, including all 8 `media.spec.ts` cases. `src/media/range.test.ts` ran 31 tests by name under `--reporter=verbose`.

Non-vacuity check on this PR's own change: neutering the seed threading in `test/browser/media.ts` to `localReader(TAB_SEED)` makes `two open tabs each stream from their own broker` fail at `media.spec.ts:201` on `expect(fromB.bodyHex)` — tab B served tab A's bytes. Restored, green again. The distinct-seed assertion binds.

Issue #1057, opened 2026-08-04
engine: mirror the ascent-link cross-check on the produce side of a re-seal


FSM1:
## Problem

The gate's stage 3 has two halves. `authenticate_section_structures` authenticates every structure signature, and `crates/engine/src/gate/adoption.rs` then opens the ascent link under the reader's cached ancestor node seed and fail-closes when `payload.epoch != envelope.epoch` or the derived read key does not match this node's.

`crates/engine/src/net/author.rs::check_scope_root` mirrors only the first half. `EnvelopeAuthoring` carries no ancestor node seed, so the second half is unavailable there — but the re-seal path does hold one (`identity.parent_node_seed` in `crates/engine/src/rotation/reseal.rs`), and it is the path that mints the link. A re-seal that sealed the wrong override seed or the wrong epoch into an interior scope root's ascent link would author and publish a record every ancestor reader rejects whole-record: exactly the unopenable-record class AGENTS.md rule 8 exists to prevent.

The module header in `net/author.rs` now states the boundary rather than implying full stage-3 parity, so this is a known gap, not a silent one.

## Acceptance

Mirror the ascent-link cross-check release-active wherever the minting path holds the ancestor node seed — most likely inside the re-seal rather than in `check_scope_root`, since that is where the seed lives. Cover it with a reject row that fires in a release build: a link sealed at the wrong epoch, and one whose override seed does not derive the node's read key.

Deferred from the #1045/#1048 PR: `crates/engine/src/rotation/*` was owned by a sibling PR running in parallel.

Gate: `Engine Tests`.

Part of the v2 build.

Issue #1056, opened 2026-08-04
engine: surface the produce-side refusal check name instead of only attempts-exhausted


FSM1:
## Problem

`AuthorError::check` gives every produce-side refusal a stable kebab-case name, the counterpart of `GateRejection::check`, and `Display` renders it. No production consumer reads either:

- `crates/engine/src/sync/drain.rs` maps the error through `classify_author`, which consumes it and returns a `Halt`;
- `crates/engine/src/net/rotation.rs` maps it to `ScopeRootPublishError::NotPublished` with `|_|`.

So a produce-side trust refusal surfaces only as `DeadLetterReason::AttemptsExhausted` five drain passes later, indistinguishable from a network outage. That is the trust-vs-availability conflation the read side deliberately avoids: the gate's own rejections carry their check name all the way to `emit_trust_violation`.

## Acceptance

Carry the check name from the authoring refusal to the host: either widen the drain's dead-letter reporting so the op's notice names the refusal, or emit it on the event stream the way a gate rejection is emitted. Assert it end-to-end — a drain pass that refuses to author reports which check fired, not merely that attempts ran out.

Deferred from the #1045/#1048 PR: the facade's event surface was owned by a sibling PR running in parallel.

Gate: `Engine Tests`.

Part of the v2 build.

Issue #1055, opened 2026-08-04
engine: classify rotation-path authoring refusals instead of flattening them to not-published


FSM1:
## Problem

`crates/engine/src/net/rotation.rs` maps every `AuthorError` from `author_scope_root_with_section` to `ScopeRootPublishError::NotPublished`, which `is_retryable` reports as retryable. The four produce-side trust refusals are not: they are the adoption gate's own verdict reached before the PUT, and re-authoring the same section reaches the same verdict. A rotation whose section cannot author therefore retries forever without ever being classified.

`crates/engine/src/net/author.rs` now carries the classification the drain uses — `AuthorError::is_trust_refusal` and `AuthorError::check` — so the mapping is a one-line routing change plus the verdict decision:

```rust
.map_err(|e| if e.is_trust_refusal() {
    ScopeRootPublishError::Rejected
} else {
    ScopeRootPublishError::NotPublished
})?;
```

`Rejected` is the non-retryable arm the rotation plane already uses for a failed commitment verify, so the shapes line up.

Two things to settle while there:

1. `check_publishable` verifies the commitment signature at `net/rotation.rs:339-342`, and `check_scope_root` now verifies it again inside the authoring call nine lines later. One of the two should go. Deleting the pre-check is the tidier direction, but it moves the verdict from `Rejected` to whatever the `AuthorError` mapping produces — so it only becomes safe once item 1 above lands. Do not delete it first.
2. The sweep and cascade paths consume `ScopeRootPublishError` through `is_retryable`; check that a newly non-retryable verdict there drops the node from the work list rather than aborting a cascade mid-way. The denial-of-revocation shape from #1047 is the thing to avoid: a committed grantee must never be able to make an owner's rotation fail permanently.

Deferred from the #1045/#1048 PR: `net/rotation.rs` was owned by a sibling PR running in parallel, so that PR made only the one-argument call-site change the new `check_scope_root` signature forced.

Gate: `Engine Tests`.

Part of the v2 build.

PR #1058, opened 2026-08-04
fix: mark an op published at its record ack and mirror the gate stage 2 on the produce side

### LINKED ISSUE #1045: engine: a publish whose record acked but whose self-adopt failed is still replayable
The #904 PR closes the replay window for a published op with a durable `PUBLISHED_OP_MARK_KEY` high-water, raised between the op  s last record publish and the release of its staged blocks. It covers what the mark claims: publish **and self-adopt** confirmed.

One window stays open.

## The window

`Drain::publish_node` (`crates/engine/src/sync/drain.rs`) returns `Ok` only once the self-adopt through `RootAdopter`/`ChildAdopter` has also succeeded. `publish_head` before it returns `Ok` on `PublishOutcome::Published` — the record is live at its name and names the version  s content CIDs. If the adopt then fails, `publish_node` returns `Err`, the mark is never raised, and the op stays queued.

On the next boot the op replays, re-uploads the identical content-addressed blocks, they enter `confirmed`, and a cancel arriving during that transfer reaches `retire_cancelled` — unpinning content a live record names. `cancel_upload`  s new durable guard does not fire, because the mark is low.

Within one session the session-scoped `past_publish_entry` interlock still covers it; across a reboot it does not, which is the same shape as #966.

## Resolving

Raise the mark on the confirmed publish rather than on the confirmed adopt — inside `publish_head`  s `PublishOutcome::Published` arm. That needs the op id threaded through `publish_node` into `publish_head`, and a decision about folder publishes, which carry no op id of their own. Note `PublishOutcome::Unconfirmed` must **not** raise it: those bytes may never have landed, and dropping the op would be loss.

Until then the doc comments on `PUBLISHED_OP_MARK_KEY` and `Drain::mark_published` state the narrower guarantee the code actually gives.

Gate: `Engine Tests`.

Surfaced by the `/crypto-privacy-review` gate on the #904 PR.

Part of #655

### LINKED ISSUE #1048: engine: classify produce-side trust refusals and complete the author-side stage-2 mirror
## Problem

Two gaps in the publish side's fail-closed mirror, both surfaced by the review gates on the write-body/history-link epoch fix.

### 1. A permanent produce-side invariant violation is classified as a transient outage

`publish_node` in `crates/engine/src/sync/drain.rs` maps every `AuthorError` to `Halt::Unclassified`, which is documented as charged-nothing-and-retried, so an outage never abandons an op. But `SectionSignatureInvalid`, `CommitmentNameMismatch`, `MissingGrantSection` and `GrantSectionOnChild` are deterministically permanent: the same bytes fail on every retry, so the op spins forever, uncharged against the attempt budget and never dead-lettered.

Not reachable on the normal drain path today — `pass.epoch` and the carried `envelope_unknown` both come from the same record in `load_scope_root`, so they agree by construction — but the classification is wrong in kind. `crates/engine/src/net/rotation.rs` collapses the same errors into `ScopeRootPublishError::NotPublished`, which has the same shape of problem.

**Acceptance:** match on `AuthorError` at the drain and rotation call sites and route the section/commitment invariant violations to a permanent verdict, leaving only genuinely transient causes retryable. Give the produce-side trust refusals stable check names the way `GateRejection::check` does, so an epoch-pairing regression is visible in the field rather than silent.

### 2. The produce-side mirror covers gate stage 3 but not stage 2's signature check

`check_scope_root` verifies the commitment's `ipnsName` binding and every structure signature, but never verifies `commitment_sig`, the owner ECDSA signature the gate's stage 2 checks. `authenticate_section_structures` is self-referential: it authenticates structures against the pseudonyms the section's own commitment names. So a wholly self-consistent attacker-authored section — own commitment, own pseudonym, own signatures, correct `ipnsName` — passes `check_scope_root` completely.

The carried section on the drain path comes from the snapshot cache, which is floor-checked but does not re-verify the section's signatures, so a tampered cache entry is caught for internal consistency but not for owner authority. The post-publish self-adopt runs the full gate and would catch it, but only after a sequence number is burned. `net/rotation.rs::check_publishable` already verifies the commitment signature on the rotation publish path; the drain path has no equivalent.

**Acceptance:** thread the owner `EcdsaVerifier` into `check_scope_root` and call `verify_grant_set` alongside the stage-3 call, or document explicitly why stage 2's signature half is deliberately outside the produce mirror.

Part of the v2 build.

FSM1:
Two gaps in the write plane's fail-closed publish path, both extending what landed tonight in #1046 and #1049.

## The mark's line is the ack, not the self-adopt

`publish_node` returns `Ok` only once the self-adopt has also succeeded, so a record whose PUT confirmed and whose adopt then failed left the op queued with the published-op mark unraised. On the next boot that op replayed, re-uploaded the identical content-addressed leaves, and a cancel landing in that transfer reached `retire_cancelled` and unpinned content the live record already names. Within one session the publish-entry interlock covers it; across a reboot `UploadCancels` is empty, so it does not.

The mark now rises inside `publish_node` the instant `publish_head` returns, which it does only on `PublishOutcome::Published`. An unconfirmed or race-losing publish raises nothing: those bytes may never have landed, and dropping the op on the strength of that would be loss.

The decision the issue flagged as open is which record of a plan marks. It is the plan's **last**: `updateContent` marks on the target's own publish, `create` on the parent-folder publish, and every other publish passes `None`. Marking the create's child publish would drop an op on restart whose child no parent names, so the negative direction is pinned by its own test.

## The produce mirror covered stage 3 but not stage 2

`check_scope_root` verified the commitment's `ipnsName` binding and every structure signature, but never `commitment_sig`. `authenticate_section_structures` authenticates structures against the pseudonyms the section's **own** commitment names, so a wholly self-consistent attacker-authored section — own commitment, own pseudonym, own signatures, correct `ipnsName` — passed it completely. The drain reads its scope root from the snapshot cache, which is floor-checked but never re-verifies the section, so a tampered entry was caught for internal consistency and not for owner authority.

`check_scope_root` now runs `verify_grant_set` at the gate's own stage-2 position, release-active, on both scope-root authoring paths. Reject rows for a foreign-signed commitment, an unusable signature, and the re-seal path's own equivalent, all returning `Err` so they fire in a release build.

Produce-side refusals also carry stable check names — `AuthorError::check`, rendered by `Display` — and the drain charges the trust class against the attempt budget instead of retrying it free forever.

## What differs from the issue bodies

- `#1048` asks for a permanent verdict. These refusals are charged instead. On the drain path they can only come from the cached scope root, which a later resolve replaces, so an immediate dead-letter would abandon a user's ops over a cache another tick repairs — the mirror-image of the bug being fixed. The budget bounds the spin either way, and `Seal`/`HeadTooLarge` stay availability because a rebase may not build that body again.
- `#1048`'s second half also names `net/rotation.rs`. That file is a sibling PR's, so this branch makes only the one-argument call-site change the new `check_scope_root` signature forces; the classification there is #1055.
- `#1045`'s `MissingGrantSection`/`GrantSectionOnChild` are unreachable from the drain today — a gated child never carries a section and a gated root always does — but they are classified with the rest rather than left as the one uncharged arm.

## Also folded in from the review gates

A reload that lands on a rotated root is now refused rather than authored at the epoch the pass opened on: that skew is a self-healing race, and the new classification would otherwise spend the op's budget on it. Two invariants the trust class made stale are corrected — an authoring refusal is raised before its own record reaches the transport, not before every PUT of the op, and the abandonment retires on the target still being unreachable rather than on no PUT having acked.

## Deferred

`#1055` rotation-path classification, `#1056` surfacing the refusal name past `AttemptsExhausted`, `#1057` the ascent-link half of stage 3 on the re-seal's produce side. Each carries a native dependency edge on `#1048`.

Closes #1045
Closes #1048

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of scope-root signatures and owner identities.
  * Added clearer handling for invalid or unauthorized commitment signatures.
  * Prevented replaying operations after confirmed publication, including after crashes or restarts.
  * Preserved queued operations when final publication does not complete.
  * Improved handling of moves, deletions, renames, and partial publication failures.
  * Improved classification of authoring refusals during publishing.

* **Tests**
  * Added coverage for signature mismatches, malformed sections, publication recovery, replay prevention, and publication failures.

> [!NOTE]
> ### Mark ops published at record PUT ack and verify commitment signatures on scope-root authoring
> - Published-op marks are now raised at the instant the last record's PUT confirms (in `publish_node`), rather than after self-adopt; this improves crash safety and prevents redundant re-uploads on restart.
> - Scope-root authoring in `check_scope_root` now performs stage-2/3 validation: decodes the grant section, verifies the commitment signature against the owner identity, checks `ipns_name` match, and authenticates structure signatures. Two new `AuthorError` variants (`InvalidGrantSection`, `CommitmentSignatureInvalid`) surface these failures.
> - Cross-folder move compensation in `publish_ref_move` no longer rolls back a dest-add if the source-remove already confirmed; compensation is gated on the `PublishHalt.confirmed` flag introduced by the new `PublishHalt` struct.
> - Mid-pass scope-root rotation is now detected in `reload_folder`: if the loaded epoch differs from the pass epoch, the pass halts and retries rather than authoring with a stale epoch.
> - Risk: ops that previously marked after self-adopt now mark earlier (on PUT ack); any crash between PUT ack and self-adopt no longer causes a replay, but callers that expected the old ordering may observe the mark sooner.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 91cdac1.</sup>
> 
>

### DIFF
--- a/crates/engine/src/net/author.rs
+++ b/crates/engine/src/net/author.rs
@@ -9,9 +9,10 @@
 //! - a child envelope carrying a grant section returns `Err`, off core's own
 //!   `has_grant_section` — the produce guard and the decode reject cannot drift;
 //! - a scope root missing its grant section, carrying one whose commitment
-//!   names another `ipnsName`, or carrying one whose structure signatures do not
-//!   recompute at the envelope's own scope/epoch, returns `Err` off the same
-//!   decoder and the same stage-3 predicate the gate runs;
+//!   names another `ipnsName` or is not signed by the anchored owner identity,
+//!   or carrying one whose structure signatures do not recompute at the
+//!   envelope's own scope/epoch, returns `Err` off the same decoder and the same
+//!   stage-2/stage-3 predicates the gate runs;
 //! - a kind transplant and a non-canonical child-ref `ipnsName` are
 //!   unrepresentable: [`new_child`] feeds one [`NodeKind`] and one typed
 //!   [`IpnsName`] to both the body and the parent's ref.
@@ -21,8 +22,9 @@ use cipherbox_core::ipns::IpnsName;
 use cipherbox_core::seal::{
     ChildRef, Envelope, GrantSection, NodeKind, PreservedFields, ReadBody, Version,
     decode_grant_section, encode_envelope, encode_grant_section, grant_section_bytes,
-    has_grant_section, seal_read_body, set_grant_section,
+    has_grant_section, seal_read_body, set_grant_section, verify_grant_set,
 };
+use cipherbox_core::suite::ecdsa::{EcdsaSignature, EcdsaVerifier};
 
 use crate::content::limits::MAX_RESOLVED_RECORD_BYTES;
 use crate::content::root_block_cid;
@@ -46,6 +48,9 @@ pub enum AuthorError {
     /// The carried grant-set commitment names a different `ipnsName` than the
     /// record is published under, which the gate's stage 2 rejects.
     CommitmentNameMismatch,
+    /// The carried grant-set commitment is unsigned, malformed, or not signed by
+    /// the contact-anchored owner identity, which the gate's stage 2 rejects.
+    CommitmentSignatureInvalid,
     /// A structure signature in the carried grant section does not recompute at
     /// the envelope's own scope and epoch, which the gate's stage 3 rejects
     /// whole-record.
@@ -65,6 +70,45 @@ pub enum AuthorError {
     },
 }
 
+impl AuthorError {
+    /// The stable kebab-case name of this refusal — the produce-side counterpart
+    /// of [`GateRejection::check`], so a regression in the epoch or commitment
+    /// pairing is legible in the field rather than a silent retry.
+    ///
+    /// [`GateRejection::check`]: crate::gate::GateRejection::check
+    pub fn check(&self) -> &'static str {
+        match self {
+            Self::GrantSectionOnChild => "grant-section-on-child",
+            Self::MissingGrantSection => "missing-grant-section",
+            Self::CommitmentNameMismatch => "commitment-name-mismatch",
+            Self::CommitmentSignatureInvalid => "commitment-signature-invalid",
+            // The gate's own verdict name: this is the same predicate.
+            Self::SectionSignatureInvalid => "structure-signature-invalid",
+            Self::Seal(e) => e.check(),
+            Self::HeadTooLarge { .. } => "head-too-large",
+        }
+    }
+
+    /// Whether this is a **trust** refusal on the record's own bytes — the
+    /// produce-side mirror of an adoption-gate rejection — rather than a codec
+    /// or size refusal of the body this pass happened to build, which a rebase
+    /// onto other state may not reproduce.
+    ///
+    /// The write plane charges a trust refusal against its attempt budget:
+    /// re-authoring the same inputs repeats it verbatim, so leaving it
+    /// unclassified spins the queue uncharged and forever (#1048).
+    pub fn is_trust_refusal(&self) -> bool {
+        match self {
+            Self::GrantSectionOnChild
+            | Self::MissingGrantSection
+            | Self::CommitmentNameMismatch
+            | Self::CommitmentSignatureInvalid
+            | Self::SectionSignatureInvalid => true,
+            Self::Seal(_) | Self::HeadTooLarge { .. } => false,
+        }
+    }
+}
+
 impl core::fmt::Display for AuthorError {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         match self {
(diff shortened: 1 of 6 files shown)


FSM1:
> **Consider adding a `heal_put_for` counterpart.** … The current tests do not need it. Add it when a retry-completion test arrives.

Addressed in 91cdac110 — but not as a bare helper, because an unused `pub fn` in a `pub mod testkit` raises no `dead_code` warning and would have rotted silently. Adding it alone is exactly the speculative generality the `/simplify` gate strips.

The asymmetry was real, though, and it was pointing at a genuine coverage gap rather than a missing method. `a_cross_folder_moves_dest_add_never_marks_on_its_own` closes on:

```rust
"the move stays queued for the retry that completes it"
```

…and then never runs that retry. The claim was asserted, not proven.

So the retry-completion test you said to wait for is in this commit, and it is what justifies the helper:

- `crates/engine/src/testkit/fakes/record_store.rs:117` — `heal_put_for`, and `fail_put_for`'s doc now names it the way the other two fail/heal pairs do.
- `crates/engine/tests/write_plane.rs:4700` — `a_retried_cross_folder_move_marks_when_its_source_remove_lands`: fail the root's PUT, confirm the compensated attempt leaves the mark down, heal, drive one more drain pass, then assert the dest-add republishes, the mark advances to the op, and the op leaves the queue.

It passed first run, so the engine already behaved as the original test claimed — the gap was in the proof, not the code.

`cargo fmt --all --check` and `cargo clippy --workspace --all-targets` are clean; `cipherbox-engine` is 890 green, `write_plane` 89 → 90.

Issue #1064, opened 2026-08-04
web: the media player has no surface for a recoverable stream refusal


FSM1:
`EngineError::TooManyStreams` is raised in exactly one place — `Engine::open_content_stream` (`crates/engine/src/facade.rs`). The snapshot pull path (`Engine::snapshot`) never opens a content stream, so the code can never reach `SnapshotStore` and therefore never reaches the vault browser.

The web client now classifies it (`isRecoverableEngineError`, `packages/client/src/correlatedTransport.ts`) and the vault browser renders a recoverable error as a retryable notice over the last-known-good listing rather than as a terminal failure. That surface is correct but currently dormant: nothing the snapshot path can return is classified recoverable.

The reachable path is the media pipe. `MediaBroker` handles the ceiling itself — it gives back the streams held only by the pin linger and retries the refused open once — but if that retry also fails, the body errors and the media element sees a bare network failure with no explanation and no affordance.

Scope: once a player component exists in `apps/web`, give it an error surface that distinguishes a recoverable ceiling refusal from a fault, fed by the engine code rather than by the message string. That needs `cb:media:error` to carry the engine's code across the port (`packages/client/src/media/protocol.ts`), which it deliberately does not today because nothing consumes it.

Surfaced by the review gates on the media-pipe slice.

Issue #1063, opened 2026-08-04
client: an idle media body cannot be ended until its next pull times out


FSM1:
`MediaPipe` keeps one `ResponseSink` per request only between a pull and its answer (`packages/client/src/sw/pipe.ts`), and the body stream runs at `highWaterMark: 0`. A media element that has buffered enough therefore has no pull in flight, and in that window the pipe has no way to deliver anything to the body.

Two consequences, both latency rather than correctness:

- A `cb:media:error` the tab posts unsolicited — as `MediaBroker.revoke` now does when a stream URL is withdrawn — is dropped. The body only fails on its next pull, which finds no cursor and then dies on the 30 s `pullTimeoutMs`.
- `detachPort` posts `cb:media:close` for every live body on a dying port, which correctly releases the tab-side cursor and its pin, but a body that was idle at that moment still stalls for the full pull deadline before it errors. That deadline also calls `discardPort`, taking every sibling body on the same port down with it.

Fix shape: capture the `ReadableStreamDefaultController` when the body is constructed and hold it alongside the port in the `bodies` map, so `detachPort` and an unsolicited error can end an idle body immediately instead of waiting for a pull that will time out.

Surfaced by the review gates on the media-pipe slice.

Issue #1062, opened 2026-08-04
client: the media head frames a range from the ticket mint-time size, not the version it pins


FSM1:
`MediaBroker.open` frames `Content-Range` and `Content-Length` from the ticket's mint-time `source.size` (`packages/client/src/media/registry.ts` `MediaSource`), but the engine stream is opened lazily on the first pull and pins whatever the head version is at that moment. The head therefore describes one version's length while the body carries another's bytes.

Per-ticket pinning made this consistent across a playback rather than per-request, which is as far as the client can take it: `EngineHandle.openContentStream` (`crates/wasm/src/host.rs`) resolves with a bare handle id, so the tab has no way to learn the size of the version it just pinned.

The shrink direction is covered — a short read ends the body as a clean EOF (`packages/client/src/media/broker.ts`) — at the cost of under-delivering `Content-Length`, which a media element reads as a network error. Framing from the pinned version would make the head exact instead.

Fix shape: have `open_content_stream` report the pinned `Version`'s size across the wasm boundary, and frame the head from it rather than from the registry entry.

Surfaced by the review gates on the media-pipe slice.

Issue #1006, opened 2026-08-03
web: the stream-ceiling refusal reads as a generic failure instead of a recoverable one


FSM1:
The engine bounds its live-stream table at `MAX_OPEN_STREAMS` and refuses past it with `EngineError::TooManyStreams` (`crates/engine/src/facade.rs:2495`, `:2506`). `crates/wasm/src/host.rs:482` surfaces that to JS as the code `tooManyStreams`.

Nothing on the web side knows the code. `EngineRequestError` carries `.code` (`packages/client/src/correlatedTransport.ts:35`) and `apps/web/src/engine/snapshotStore.ts:154` passes it through, but no view branches on a code, so the refusal renders as the raw `Display` string like any other engine failure.

That is wrong in kind, not just in wording. `tooManyStreams` is **recoverable** — closing an idle stream frees a slot and the same call succeeds — whereas the neighbouring `unknownStreamHandle` is terminal. The UI presents them identically, so a user who hits the ceiling with several media elements open sees a dead-end error where a retry would have worked. `MediaBroker` compounds it: any throw posts `cb:media:error` and drops the cursor, so playback ends rather than backing off.

Desktop already handles this correctly — `crates/fuse/src/error.rs:71` maps it to `VfsError::Unavailable`, which the kernel retries. Only the web client lacks the affordance.

Scope:

- Recognise `tooManyStreams` in the web error surface and render it as transient with a retry, not a failure.
- Decide whether `MediaBroker` should release an idle cursor and retry once rather than dropping the cursor outright.

The refusal itself is correct and fail-closed; this is purely the client-side presentation and recovery of it. Surfaced by the review gates on #980.

Part of #993

FSM1:
Scope bullet 2 is answered in #1065: `MediaBroker` gives back the streams held by its pin linger alone and retries the refused open once, rather than dropping the cursor outright. Per-ticket stream pinning in the same PR is the larger half of the same fix, since a page issuing many concurrent Range requests for one file used to cost one engine stream each.

Scope bullet 1 is still open, and this issue's diagnosis of where it lives needs correcting. The body points at `apps/web/src/engine/snapshotStore.ts` as the surface to branch on, but that surface can never receive the code:

- `EngineError::TooManyStreams` is constructed at exactly two sites, `crates/engine/src/facade.rs:2585` and `:2594`, both inside `open_content_stream`.
[...]

PR #1065, opened 2026-08-04
fix: pin one engine stream per media ticket and surface the stream ceiling as recoverable

### LINKED ISSUE #1007: client: the media pipe pins one engine stream per Range request instead of per ticket
`MediaBroker` opens an engine stream per intercepted fetch rather than per playback ticket. `sw/pipe.ts` mints a fresh `requestId` for every intercepted Range request, and `media/broker.ts:151` opens the stream lazily on the first pull of that request, so a `<video>` re-pays `head_version` plus the full root-manifest fetch on every Range request and every seek. On a 4 GiB file that is roughly 150 KB of manifest re-downloaded per seek.

Pinning per **ticket** with a refcount, and releasing when the last cursor for that ticket drops, removes the repeated resolve and manifest fetch.

It also gives a strictly stronger form of the #948 invariant. Today the version is pinned for the life of one *response*; per-ticket pinning holds one version across a whole *playback*, which is what a viewer actually experiences as a single read.

Related residual noted during review: `MediaBroker.open` frames `Content-Range`/`Content-Length` from the ticket's `source.size`, but the stream is opened lazily and pins whatever the head is at first pull. A version change between ticket mint and first pull still frames one version's bytes under another version's headers. The short-read EOF at `broker.ts:165` covers the shrink direction only. Per-ticket pinning closes this too.

Surfaced by the review gates on #980.

### LINKED ISSUE #1010: client: bind MEDIA_WINDOW_BYTES to the engine chunk size instead of hand-copying it
`MEDIA_WINDOW_BYTES` (`packages/client/src/media/protocol.ts`) must equal `ContentProfile::PRODUCTION.chunk_size` (`crates/engine/src/content/profile.rs`) or a window at a chunk-aligned offset stops covering exactly one leaf and every read fetches two. Today the TS side is a hand-copied literal and the only thing binding the two is a prose cross-reference in a doc comment.

That is not hypothetical drift. Before #980 the TS constant was `1024 * 1024` while the engine chunk is `1_048_536` — 1 MiB less core's seal overhead, because the 1 MiB budget belongs to the sealed block, not the plaintext. Every media window on `main` therefore straddles two leaves. #980 corrects the value; nothing stops it drifting again.

The value looks like a typo for 1 MiB, so it invites well-meaning "fixes" — the CodeRabbit CLI review on #980 flagged it as major and proposed reverting to `1_048_576`. A guard needs to fail loudly rather than rely on the comment being read.

Options: expose the profile chunk size across the WASM boundary and derive the constant, or assert equality in a cross-language KAT so a mismatch fails CI. The engine profile is frozen by the wire format, so a test that pins both sides is likely enough.

Surfaced by the review gates on #980.

FSM1:
Closes #1007
Closes #1010
Part of #1006
Part of #981

Entirely within `packages/client`, plus one line of `.github/workflows/ci.yml`. `apps/web` is untouched — see "What was cut" below.

## The pin

`MediaBroker` opened an engine stream per intercepted Range request, so a `<video>` re-paid `head_version` plus the full root-manifest fetch on every seek. Streams are now pinned per **ticket**, refcounted across the cursors reading them.

A `<video>` closes a body before opening the range it seeks to, so a strict refcount would release on every seek and buy nothing. A pin whose count reaches zero is therefore held for `DEFAULT_PIN_LINGER_MS` before its stream is released. That linger is what makes the pin span a playback rather than a response, which is the stronger form of the #948 invariant #1007 asks for — and it is also the reclaimable slack the ceiling recovery below spends.

Sharing one handle across cursors needed three guards that per-request streams did not:

- A body is **bound** to the stream it read its first window from. If the ticket re-opens under it, the body fails rather than concatenating two content versions under one `Content-Range`. Reachable today: `EngineClient.retireHandles` clears the stream table on every leadership change, so one cursor noticing `unknownStreamHandle` and re-opening the ticket would otherwise splice a live sibling body.
- A failed **open** is not memoized, or every cursor on the ticket replays it. A failed **read** leaves the pin alone — only a handle the engine has provably forgotten drops it.
- The recoverable-open retry re-checks that its pin is still in `pins` across the awaits. Every release path reaches a stream through that map, so opening onto a dropped pin would strand the handle and its content key for the life of the worker.

`StreamRegistry.register` now refuses a ticket it already holds. Reuse was benign when each request opened from its own `source.node`; with a pin keyed by ticket it would serve one file's stream under another file's head.

## Reclaiming a dead port's cursor

`MediaBroker.drop` was the only path that closed an engine stream, and `MessagePort` has no close event, so a Service Worker that lost its port left the tab holding the cursor and the stream it pinned. `MediaPipe` now tracks every response body still streaming and the port it reads from — a `ResponseSink` exists only between a pull and its answer, so it cannot serve as that record — and `detachPort` posts `cb:media:close` for each of them **before** it disentangles the port. An `open` the pipe gave up waiting for, and a head that carries no body, post one too.

`MediaPipe.respond` also builds a body only for 200 and 206. The port that names the status is untrusted, and a null-body status reaching `Response` throws after the tab has already opened a cursor.

## Withdrawing a ticket

A `/stream/<ticket>` URL is a bearer capability to one file's plaintext. `revokeStreamUrl` deleted the registry entry but said nothing to the broker, so a body already streaming read on to EOF. It now ends those bodies and releases the ticket's engine stream, so revoking withdraws the capability from the bytes in flight, not merely from the next request.

## The window constant

`MEDIA_WINDOW_BYTES` must equal `ContentProfile::PRODUCTION.chunk_size` or every chunk-aligned read straddles two leaves. `protocol.test.ts` pins it against `crates/engine/kat/manifest.json`, which `crates/engine/tests/kat_content.rs` already pins against the profile — so the chain is gated at both ends. The `ts` paths filter gained `crates/engine/kat/**`, without which a Rust-side chunk-size change would skip the suite that catches it.

## The stream ceiling, and what was cut

`tooManyStreams` is a ceiling, not a verdict: closing a stream makes the same call succeed. `isRecoverableEngineError` names it, and the broker acts on it — it gives back the streams held by the pin linger alone and retries the refused open once. The ceiling recovery is exactly the slack the linger creates. Per-ticket pinning is the larger part of the same fix: a page issuing many concurrent Range requests for one file used to cost one engine stream each.

**An earlier revision of this PR also added a retryable-notice surface to the vault browser, and it has been removed. Please do not re-add it.** #1006's body points at `apps/web/src/engine/snapshotStore.ts` as the place to branch on the code, but that surface can never see it:

- `EngineError::TooManyStreams` is constructed at exactly two sites, `crates/engine/src/facade.rs:2585` and `:2594`, both inside `open_content_stream`.
- `open_content_stream` has exactly one caller in the whole Rust tree — the wasm binding at `crates/wasm/src/host.rs:392`.
- Every other TypeScript reference to `openContentStream` is transport plumbing or a test double; the only site that originates one is `packages/client/src/media/broker.ts`.
- `apps/web/src` contains no reference to `openContentStream`, and no download or preview path at all.
- `snapshotStore`'s error handling is reached only from the `facade.snapshot()` and `facade.setFocus()` rejections, neither of which opens a content stream.

So the notice, its retry button, `SnapshotStore.retry`, the `hasListing` derivation and the CSS were a branch on a code that surface cannot produce — speculative generality by `AGENTS.md`'s own definition. `isRecoverableEngineError` and `engineErrorCode` stay, next to the error they classify, because the broker genuinely uses both; the barrel export in `index.ts` went with the cut, since no consumer outside the package has one yet.

#1006 therefore stays open on its first scope bullet, and #1064 — the player surface that can actually receive the code — is now registered as blocking it. Its second bullet, whether `MediaBroker` should reclaim and retry rather than drop the cursor outright, is answered here: it reclaims and retries once.

## What remains of #981

- **Item 1, the engine cap: already done on `main`.** `MAX_OPEN_STREAMS = 256` and `EngineError::TooManyStreams` at `crates/engine/src/facade.rs`, surfaced as `tooManyStreams` by `crates/wasm/src/host.rs`, with the N+1'th open refused without disturbing live handles at `crates/engine/tests/write_plane.rs`. Not this PR's work.
- **Item 3, the window parity gate: closed here.**
- **Item 2, the dead port: closed for every case in which the worker is alive to run code** — a replaced port, a port discarded by the pull deadline, and an abandoned open all now release the tab's cursor and its pin immediately.

  The residual is the sub-case the issue names second: a worker **killed outright** runs no code, so nothing posts the close, and its cursors survive until the tab re-brokers or disposes — which `MediaService.rebroker` to `broker.serve` to `close()` already does, unchanged by this PR. The issue's other option, a per-cursor idle timer, was considered and rejected: nothing in the protocol distinguishes a paused `<video>` — which legitimately holds a body open without pulling for as long as the user likes — from a dead port, so a timer would drop live cursors and convert a resume into a stalled body. Closing this properly wants a signal, not a heuristic, so #981 stays open on that one point.

Separately, `MediaBroker.revoke` and `detachPort` both release the cursor promptly, but a body that is *idle* at that moment has no sink to receive the news and only settles on the 30 s pull deadline. Latency, not correctness — the cursor and pin are already gone. Tracked in #1063.

## Follow-ups filed

All three are sub-issues of #993 so they cannot orphan.

- #1062 — the media head still frames a range from the ticket's mint-time size rather than the version it pins; closing it needs `open_content_stream` to report the pinned version's size across the wasm boundary.
- #1063 — an idle media body cannot be ended until its next pull times out.
- #1064 — the web media player has no surface for a recoverable stream refusal. Registered as blocking #1006.

## Gates

`pnpm install`, `pnpm typecheck` and `pnpm lint` all clean. `pnpm --filter @cipherbox/client test --run` is 368 passed / 368; `pnpm --filter @cipherbox/web test --run` is 110 passed / 110, unchanged from `main` now that `apps/web` carries no diff.

No Rust source is touched, so the cargo gates are unaffected. The only file outside `packages/client/src` is the `ts` paths filter in `.github/workflows/ci.yml`, which the #1010 guard needs in order to fire at all.

`/simplify`, `/security-review` and `/crypto-privacy-review` all ran on this diff. Their findings — a stranded engine stream and never-zeroized content key on the retry path, a content-version tear via a retired handle, an un-wiped plaintext window on a failed transfer, and a poisoned pump chain — are folded in, and the gates were re-run on the fixes.

## Summary by CodeRabbit

- **New Features**
  - Improved media streaming with shared stream reuse, configurable read windows, and recovery from temporary capacity limits.
  - Added safer handling for stream revocation, cancellation, port replacement, and interrupted downloads.

- **Bug Fixes**
  - Prevented duplicate media tickets from being reused.
  - Fixed stale stream and cursor cleanup after failed, timed-out, or abandoned requests.
  - Improved recovery when stream connections close unexpectedly.
  - Ensured failed or abandoned transfers do not retain sensitive data.
  - Improved validation of revoked stream URLs and error responses.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,9 @@ jobs:
             ts:
               - 'apps/**'
               - 'packages/**'
+              # The TS suite pins MEDIA_WINDOW_BYTES against this manifest, so a
+              # Rust-side chunk-size change must run it.
+              - 'crates/engine/kat/**'
               - 'pnpm-lock.yaml'
               - 'pnpm-workspace.yaml'
               - 'package.json'
--- a/apps/web/src/components/file-browser/FileBrowser.tsx
+++ b/apps/web/src/components/file-browser/FileBrowser.tsx
@@ -5,33 +5,18 @@ import { FileList } from './FileList';
 
 /** The vault browser: where you are, what is in it, and how to move. */
 export function FileBrowser() {
-  const { rows, breadcrumbs, isLoading, isRoot, hasListing, error, retry, navigateTo, navigateUp } =
+  const { rows, breadcrumbs, isLoading, isRoot, error, navigateTo, navigateUp } =
     useFolderNavigation();
-  // A retryable refusal is a backoff, not a verdict on the folder, so a listing
-  // already on screen stays under it.
-  const settled = error === null ? !isLoading : retry !== null && hasListing;
+  const settled = !isLoading && error === null;
 
   return (
     <div className="file-browser" data-testid="file-browser">
       <Breadcrumbs crumbs={breadcrumbs} onNavigate={navigateTo} />
-      {error &&
-        (retry === null ? (
-          <p className="file-browser-error" role="alert" data-testid="file-browser-error">
-            {error.message}
-          </p>
-        ) : (
-          <p className="file-browser-notice" role="status" data-testid="file-browser-notice">
-            {error.message}
-            <button
-              type="button"
-              className="terminal-btn"
-              onClick={retry}
-              data-testid="file-browser-retry"
-            >
-              retry
-            </button>
-          </p>
-        ))}
+      {error && (
+        <p className="file-browser-error" role="alert" data-testid="file-browser-error">
+          {error.message}
+        </p>
+      )}
       {isLoading && (
         <p className="file-browser-loading" data-testid="file-browser-loading">
           {'// LOADING VAULT...'}
--- a/apps/web/src/components/file-browser/FileBrowser.tsx
+++ b/apps/web/src/components/file-browser/FileBrowser.tsx
@@ -5,18 +5,37 @@ import { FileList } from './FileList';
 
 /** The vault browser: where you are, what is in it, and how to move. */
 export function FileBrowser() {
-  const { rows, breadcrumbs, isLoading, isRoot, error, navigateTo, navigateUp } =
-    useFolderNavigation();
-  const settled = !isLoading && error === null;
+  const {
+    rows,
+    breadcrumbs,
+    isLoading,
+    isRoot,
+    error,
+    isRecoverable,
+    navigateTo,
+    navigateUp,
+    retry,
+  } = useFolderNavigation();
+  // A recoverable refusal is a backoff, not a verdict on the folder, so the
+  // last-known-good listing stays on screen underneath it.
+  const settled = !isLoading && (error === null || isRecoverable);
 
   return (
     <div className="file-browser" data-testid="file-browser">
       <Breadcrumbs crumbs={breadcrumbs} onNavigate={navigateTo} />
-      {error && (
-        <p className="file-browser-error" role="alert" data-testid="file-browser-error">
-          {error.message}
-        </p>
-      )}
+      {error &&
+        (isRecoverable ? (
+          <p className="file-browser-notice" role="status" data-testid="file-browser-notice">
+            {error.message}
+            <button type="button" onClick={retry} data-testid="file-browser-retry">
+              {'RETRY'}
+            </button>
+          </p>
+        ) : (
+          <p className="file-browser-error" role="alert" data-testid="file-browser-error">
+            {error.message}
+          </p>
+        ))}
       {isLoading && (
         <p className="file-browser-loading" data-testid="file-browser-loading">
           {'// LOADING VAULT...'}
--- a/apps/web/src/components/file-browser/FileBrowser.tsx
+++ b/apps/web/src/components/file-browser/FileBrowser.tsx
@@ -5,35 +5,31 @@ import { FileList } from './FileList';
 
 /** The vault browser: where you are, what is in it, and how to move. */
 export function FileBrowser() {
-  const {
-    rows,
-    breadcrumbs,
-    isLoading,
-    isRoot,
-    error,
-    isRecoverable,
-    navigateTo,
-    navigateUp,
-    retry,
-  } = useFolderNavigation();
-  // A recoverable refusal is a backoff, not a verdict on the folder, so the
-  // last-known-good listing stays on screen underneath it.
-  const settled = !isLoading && (error === null || isRecoverable);
+  const { rows, breadcrumbs, isLoading, isRoot, hasListing, error, retry, navigateTo, navigateUp } =
+    useFolderNavigation();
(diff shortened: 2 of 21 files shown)


FSM1:
@coderabbitai review

FSM1:
@coderabbitai full review

FSM1:
@coderabbitai full review

FSM1:
@coderabbitai full review

Issue #1018, opened 2026-08-03
engine: consume the scope-exit rotation triggers


FSM1:
## Already landed — do not re-implement

`rotate_scope` and the grantee-flat rotation path are written and simulation-tested in `crates/engine/src/rotation/rotate.rs`; `RotationTrigger::ScopeExit` already exists in `crates/engine/src/rotation/trigger.rs`. **Neither the primitive nor the trigger enum is in scope here.**

**The consumer landed in #1103.** Four of the five gate bullets below are delivered and tested:

- `ScopeExitRotator`, `ScopeExitReport` and `consume_scope_exit_triggers` — `crates/engine/src/rotation/trigger.rs`. The driver cuts one root per queued trigger, never short-circuits, and returns every failure in `ScopeExitReport::failed`.
- **Full-depth detection.** `rebase` resolves the granted source scope root by walking the source parent then its ancestors nearest-first against an injected scope-root policy — `crates/engine/src/sync/rebase.rs`. The old `from_parent` naming was the v1 one-level hole (`#26 D7`).
- **Dedupe** — one trigger per source scope root, first-seen order.
- `ScopeCrossing` replaced the `cross_scope` / `exits_granted_source` bool pair on both relocation ops, and `OpKind::Move` gained scope-crossing so a cross-scope move-with-rename queues a trigger.
- The `sync/mod.rs` deferral note was rewritten rather than deleted, because the residual below is real.

Do not re-open any of that.

## Remaining scope — one gate bullet

**The rotation is only ever exercised against a fake.** Both `ScopeExitRotator` implementors are test-only — `FakeRotator` (`crates/engine/src/rotation/trigger.rs:248`) and `RecordingRotator` (`crates/engine/tests/sync.rs:682`). Each pushes the `NodeId` into a `RefCell<Vec<_>>` and returns a hardcoded `RotationOutcome`; neither calls `rotate_scope`. The test fixture is a plain `Snapshot`, not a keyed scope hierarchy — there is no descendant to check and no commitment to compare against.

So the trust property this issue exists to prove is unasserted:

> the grantee rotation is **flat** — no descendant scope root is re-keyed, and the committed tag set is **byte-identical** across the rotation

Close that. When #1107 lands the production `ScopeExitRotator`, drive `consume_scope_exit_triggers` through it against a keyed multi-scope fixture and assert both halves. A scope exit that rotates the wrong breadth is as much a trust hole as one that never rotates.

[...]

PR #1103, opened 2026-08-05
feat: consume the scope-exit rotation triggers

FSM1:
## What this lands

The queued scope-exit triggers are no longer dropped on the floor, and the two
detection holes that would have let a trigger slip are closed.

- **Full-depth detection.** `rebase` now resolves the *granted source scope
  root* by walking the source parent and then its ancestors nearest-first
  against an injected scope-root policy (`replay`/`rebase_one` take
  `scope_roots: &[NodeId]`). It previously named `from_parent` — correct only
  when the moved node was a direct child of the scope root, which is exactly the
  v1 one-level coverage hole `blueprint/engine.md` calls out (`#26 D7`). A chain
  that reaches no listed root falls back to the snapshot root: over-rotating an
  enclosing root costs a wave, rotating nothing leaves a revokee holding a live
  seed.
- **Dedupe.** `replay` queues one trigger per source scope root, first-seen
  order — N ops leaving one scope are one rotation.
- **The consumer.** `rotation/trigger.rs` gains the `ScopeExitRotator` seam and
  `consume_scope_exit_triggers`, which cuts one root per queued trigger, never
  short-circuits, and returns every failure in `ScopeExitReport::failed` instead
  of swallowing it. The grantee-arm implementation over the live transport is
  not landed, so the seam is faked in tests — the same pattern `SweepResolver`
  and `WriteSubtreeResolver` already follow (`rotation/mod.rs`).
- **`OpKind::Move` now carries scope-crossing.** See the scope decision below.
- **`ScopeCrossing` replaces the `cross_scope` / `exits_granted_source` bool
  pair** on both relocation ops. "Exits a granted source but is not cross-scope"
  is incoherent; as a pair it was representable and a decoder would have had to
  reject it. As one value it is unrepresentable, so no encode/decode symmetry
  check is needed — the illegal state is gone from the type rather than guarded
  at both ends.
- **The drain** hands `replay` its scope root and halts a cross-scope `Move` the
  way it already halted a cross-scope `Relink`. Publishing a cross-scope
  relocation through the plain ref-move plan would carry the moved subtree into
  the destination still sealed at the source epoch.

## `OpKind::Move` — in scope, deliberately

The issue does not mention it, but `OpKind::Move` had no scope-exit field at
all, so a cross-scope move-with-rename could not queue a trigger. **This is in
scope and is fixed here.** A kernel rename journals exactly one `Move`
(`blueprint/desktop.md`), so `Move` — not `Relink` — is the desktop's entire
move surface. Leaving it blind would have kept precisely the class of silent
miss this issue exists to close: an exit that never rotates. It is an `Op`
schema change; v2 is greenfield and pre-cutover, and an op a build cannot parse
is `Retained`, never replayed and never removed (`sync/record.rs`), so the
change fails closed.

## Errors in the issue body, verified against the code

- **Line drift.** `scope_exit_triggers` is declared at `sync/rebase.rs:134`, not
  133, and populated at `:246`, not 245. The deferral note is `sync/mod.rs:14-17`,
  not just `:16`.
- **"The rebase side already detects the trigger" is not true.** The flag was
  never set by any production caller: `facade.rs:2222` was the *only*
  `Op::relink` construction outside tests and hard-coded `false, false`.
  Detection was effectively greenfield.
- **The orchestrating brief's claim that `sync/drain.rs:964` also hard-codes the
  flag is wrong.** `drain.rs` never constructs a relink op; `:1029` is a
  *pattern match* on `cross_scope: false`, not a construction. `facade.rs:2222`
  was the single hard-coded site.
- **"Retire the `sync/mod.rs:16` deferral note once the consumer exists"** — the
  note is rewritten rather than deleted, because deleting it outright would read
  as an all-clear the code does not support (see residuals).

## Known residuals — deliberately not done

- **No production `ScopeExitRotator`.** Assembling a `RotateScopePlan` needs the
  scope root's committed set, current override seed, history links and pointer
  read key, none of which the sync layer holds; the grantee-arm net that
  resolves them is not landed. This is the issue's own stated dependency. Hence
  `Part of #1018`, not `Closes`.
- **The engine is still single-scope**, so `facade.rs` authors
  `ScopeCrossing::Intra` for every relocation and no production path yet reaches
  `ExitsGrantedSource`. The scope-root policy is injected as a parameter, so the
  detection is exercised at depth 1 and depth N under a multi-scope fixture and
  gains nothing when the real scope map arrives.
- **`DrainReport` carries no trigger field.** An earlier draft added one; with no
  rotator to hand it to it was written and never read, and the facade cannot
  author a cross-scope op, so no test could reach it. Untestable plumbing was
  dropped in the review pass; `ReplayReport::scope_exit_triggers` — which the
  drain's own `replay` call produces — stays the surface.
- **`AlreadySatisfied` drops a scope-exiting op without queuing its trigger.**
  Unreachable today: a cross-scope relocation cannot publish, so gate-passing
  state can never show *our own* scope exit as already satisfied. The pass that
  lands cross-scope publishing must handle this branch, or a trigger is lost
  there.
- **Durability of the intent is the durable op queue.** A cross-scope relocation
  halts the pass (`Halt::Unclassified`), which charges nothing and leaves the op
  at the head of the queue, so every pass re-derives the trigger until it
  settles. No separate durable trigger store was added; one would have been a
  second source of truth for state the queue already holds.

## `facade.rs`

Touched minimally and only as forced by the type change: the two `Op::relink` /
`Op::move_node` construction sites now pass `ScopeCrossing::Intra`, one import,
and a stale trailing-bools comment replaced. No logic change. Flagging it since
`facade.rs` is a hot file this wave.

## Review gates

`/simplify`, `/security-review` and `/crypto-privacy-review` were run as
rigorous manual passes against `git diff main...HEAD` — the slash-command
subagents were unavailable in this environment (concurrency limit), so no gate
was skipped, none was run as a skill.

**`/simplify`** — four findings, all folded into the second commit: an unread
`DrainReport` field, an uncalled `ScopeCrossing::is_cross_scope`, a comment on
`rebase_one` whose stated rationale was about walking from the target (which the
code does not do), and two rebase unit tests the sync narratives already state
end-to-end.

**`/security-review`** — the trigger is a revocation-boundary surface, so the
questions were who can influence it and in which direction it fails.

- *Influence*: `ScopeCrossing` is authored only by the facade and rides an
  HPKE-to-self sealed, owner-tagged op record. A store co-tenant's record
  classifies as `Retained`, never `Mine`, so an attacker cannot inject a forged
  scope-exit — nor suppress one beyond the pre-existing queue-integrity residual
  `blueprint/engine.md` already documents.
- *Fallback direction*: over-rotate, never under-rotate. `Snapshot::ancestors` is
  cycle-guarded, so a malformed link cycle terminates rather than looping.
- *Applied-vs-queued*: every `OpResolution` branch was walked. Applied exits
  queue a trigger; `MoveRaceLost`, the relocation guards and `SuffixExhausted`
  queue none, and in each the exit did not happen. `AlreadySatisfied` is the one
  asymmetry, recorded above as a residual.
- *Dedupe*: collapses only equal roots. Two exits from distinct scopes that both
  hit the snapshot-root fallback do collapse — but in that state the engine knows
  neither scope, so it could not have rotated them separately either.
- *DoS*: trigger and scope-root vectors are bounded by distinct scope roots, so
  the linear `contains` scans are over single-digit collections.
- *Determinism*: no clock or RNG added; the scope-root policy enters as a
  parameter.

**`/crypto-privacy-review`** — the diff touches no primitive, no key or seal
material, and no trust-boundary read. `RotationOutcome` carries epochs only, and
`RotateError::check()` is an existing key-material-free classifier. No zeroize
ownership changed. Recorded rather than skipped.

## Gates run locally

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets`,
`cargo test -p cipherbox-engine -p cipherbox-core` (20 targets, all green),
`cargo check --workspace --all-targets`,
`cargo check -p cipherbox-wasm --target wasm32-unknown-unknown`, and
`node scripts/check-tracker-refs.mjs`.

Part of #1018

> [!NOTE]
> ### Add scope-exit rotation trigger consumption with `ScopeCrossing` classification for relocation ops
> - Introduces `ScopeCrossing` enum in [`sync/op.rs`](https://github.com/FSM1/cipher-box/pull/1103/files#diff-1446e839d1e2d23d053dc1688ed217910161cfb299565c5a86ce2372bc5c1ffe) to replace boolean flags on `Relink` and `Move` op variants, encoding intra/cross/exiting-granted-source semantics explicitly.
> - Adds `consume_scope_exit_triggers` in [`rotation/trigger.rs`](https://github.com/FSM1/cipher-box/pull/1103/files#diff-5dba1beebd4cfb61e574efe0c30c5e460dd3752b76a4b320200a7a2db6202e25) that iterates scope-root triggers in order, calling `ScopeExitRotator::rotate_on_scope_exit` for each without short-circuiting on failure.
> - Updates `replay` in [`sync/rebase.rs`](https://github.com/FSM1/cipher-box/pull/1103/files#diff-1fc181ef0406c58cf318ec1da43d8831ee841715ed4f3179e23a0b076ddc8850) to accept a scope-root policy slice and compute deduplicated scope-exit triggers via full-depth `source_scope_root` resolution.
> - Updates `Drain::drain_queue` in [`sync/drain.rs`](https://github.com/FSM1/cipher-box/pull/1103/files#diff-2dbe80269b4dfac050f51bbaedb9a173236f92de78f9951dd049c549f45f08a5) to pass scope roots to `replay` and halt with `Err(Halt::Unclassified)` on any cross-scope relocation.
> - Risk: callers of `Op::relink`, `Op::move_node`, `rebase_one`, and `replay` must now pass `ScopeCrossing` and scope-root arguments; existing call sites without updates will not compile.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8f236de.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added automatic processing for scope-exit rotations after content moves.
  * Reports completed rotations and preserves failed rotations for retry.
  * Processes multiple scope exits in order while preventing duplicate triggers.
  * Exposed rotation status and scope-crossing details through the public API.

* **Bug Fixes**
  * Improved handling of nested and cross-scope moves.
  * Correctly identifies source scopes during deep relocations.
  * Rotation failures are reported without stopping subsequent processing.

### DIFF
--- a/crates/engine/src/facade.rs
+++ b/crates/engine/src/facade.rs
@@ -59,7 +59,7 @@ use crate::sync::boot::{ColdStartError, ColdStartOutcome, ColdStartParams, cold_
 use crate::sync::cancel::UploadCancels;
 use crate::sync::drain::{Drain, DrainReport, DrainScope, published_op_mark};
 use crate::sync::model::{NodeMeta, Snapshot, collation_key};
-use crate::sync::op::{NewNode, Op, OpKind, Replaced, StagedContent};
+use crate::sync::op::{NewNode, Op, OpKind, Replaced, ScopeCrossing, StagedContent};
 use crate::sync::overlay::apply_overlay;
 use crate::sync::pointer::PointerFetch;
 use crate::sync::project::project_child_version;
@@ -2218,15 +2218,15 @@ impl<T: SeamTypes> Engine<T> {
             Command::Relink { node, new_parent } => {
                 let rendered = self.render().await?;
                 let (from_parent, base_sequence) = self.relocation_anchors(&rendered, node);
-                // trailing bools: cross_scope=false, exits_granted_source=false — intra-scope pure relink
+                // This session holds one scope, so every relocation it can form
+                // stays inside it.
                 let op = Op::relink(
                     node,
                     from_parent,
                     new_parent,
                     base_sequence,
                     authored_at,
-                    false,
-                    false,
+                    ScopeCrossing::Intra,
                 );
                 self.stage_and_notify(&op).await
             }
@@ -2253,6 +2253,7 @@ impl<T: SeamTypes> Engine<T> {
                     replacing,
                     base_sequence,
                     authored_at,
+                    ScopeCrossing::Intra,
                 );
                 self.stage_and_notify(&op).await
             }
--- a/crates/engine/src/lib.rs
+++ b/crates/engine/src/lib.rs
@@ -81,8 +81,9 @@ pub use profile::SyncTimingProfile;
 pub use rotation::{
     ChildIndexResolver, CommittedSet, EagerSet, EnumerationError, PrevEpochSeed, ResealError,
     ResealSeeds, ResealedScopeRoot, ResolveFailure, RevokeError, RevokedCommittedSet, RotateError,
-    RotateScopePlan, RotationOutcome, RotationTrigger, ScopeRootIdentity, ScopeRootPublishError,
-    ScopeRootPublisher, enumerate_eager_set, reseal_scope_root, revoke_read_grant, rotate_scope,
+    RotateScopePlan, RotationOutcome, RotationTrigger, ScopeExitReport, ScopeExitRotator,
+    ScopeRootIdentity, ScopeRootPublishError, ScopeRootPublisher, consume_scope_exit_triggers,
+    enumerate_eager_set, reseal_scope_root, revoke_read_grant, rotate_scope,
 };
 pub use seams::{SeamError, SeamResult, SeamSet, SeamTypes};
 pub use settings::{
@@ -95,7 +96,7 @@ pub use sync::{
     AppliedOp, BlockedOp, Connectivity, DeadLetterReason, DropReason, FocusTarget, FocusWindow,
     HeadReconciliation, Link, NewNode, NodeMeta, Op, OpKind, OpRecordError, OpResolution,
     PointerError, PointerFetch, RecordClass, RecordReader, RecordSeal, Repair, Replaced,
-    ReplayReport, SessionRole, Snapshot, StagedContent, TickCause, TickControl,
+    ReplayReport, ScopeCrossing, SessionRole, Snapshot, StagedContent, TickCause, TickControl,
     VaultPointerAdoption, apply_overlay, apply_repairs, classify, decode_queue, encode_op_record,
     focus_set, observed_repair, rebase_one, reconcile_head, record_content_root_cid, replay,
     resolve_vault_pointer, stage_op,
--- a/crates/engine/src/rotation/mod.rs
+++ b/crates/engine/src/rotation/mod.rs
@@ -10,7 +10,9 @@
 //!   the [`cascade`] (each eager-set descendant, fresh seed), and by the sweep.
 //! - [`rotate`] — `rotateScope`, the read-plane root cut: mint a fresh override
 //!   seed, re-seal, publish CAS, raise `minReadEpoch`, enqueue the sweep.
-//! - [`trigger`] — the read-revoke / scope-exit / manual trigger surface.
+//! - [`trigger`] — the read-revoke / scope-exit / manual trigger surface, and
+//!   the driver that turns a replay's queued scope-exit triggers into one flat
+//!   `rotateScope` per source scope root.
 //! - [`sweep`] — the lazy-wave epoch-lag convergence pass (metadata-only,
 //!   existing seed, `prev = None`) plus its idle-cadence driver and the
 //!   direct-child-scope index self-heal. It does **not** mint fresh descendant
@@ -28,8 +30,9 @@
 //!
 //! [`ChildIndexResolver`], [`CascadeResealResolver`], [`ScopeRootPublisher`] and
 //! [`WriteWavePublisher`] have production implementations over the real transport
-//! in [`crate::net::rotation`]. [`SweepResolver`] and [`WriteSubtreeResolver`]
-//! have no production implementation yet; tests fake them.
+//! in [`crate::net::rotation`]. [`SweepResolver`], [`ScopeExitRotator`] and
+//! [`WriteSubtreeResolver`] have no production implementation yet; tests fake
+//! them.
 
 pub mod cascade;
 pub mod eager_set;
@@ -59,4 +62,7 @@ pub use rotate_write::{
     build_repoint_object, derive_write_name, rotate_scope_write,
 };
 pub use sweep::{SweepError, SweepOutcome, SweepResolver, SweepTarget, run_sweep, sweep_pass};
-pub use trigger::{RevokeError, RevokedCommittedSet, RotationTrigger, revoke_read_grant};
+pub use trigger::{
+    RevokeError, RevokedCommittedSet, RotationTrigger, ScopeExitReport, ScopeExitRotator,
+    consume_scope_exit_triggers, revoke_read_grant,
+};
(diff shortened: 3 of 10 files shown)


FSM1:
Triaged all four nitpicks against the code and `AGENTS.md`. One accepted, three declined.

### 1. `sync/op.rs` — wire coverage for `ScopeCrossing::Cross` — **accepted** (2a32661d5)

Correct and worth more than "trivial". `ScopeCrossing::Cross` appeared nowhere in the repo outside the enum definition — not one construction, not one assertion. It is a serialized variant of a wire type this PR introduces, so it shipped with zero round-trip proof. Added the case to `every_op_round_trips_through_the_body_encoding`.

Proven non-vacuous: renaming the variant's serde tag to collide with `Intra` fails the new case alone (`crossing: Intra` vs `crossing: Cross`) and leaves the pre-existing cases green; restored byte-identical.

Also checked the encode/decode fail-closed symmetry rule while here. Nothing to add: `encode_body`/`decode_body` are plain serde over the enum, and there is no decode-side hard-reject for this type to mirror — replacing the `cross_scope`/`exits_granted_source` pair with one enum is precisely what removed the rejectable state, making the invariant unrepresentable rather than checked.

### 2. `facade.rs` — derive the crossing via a `crossing_for` helper — **declined**

The proposed helper ignores both parameters and returns a constant. That is not derivation, it is the same assertion behind indirection, and `/simplify` flags exactly this shape: speculative generality with one trivially-constant caller. It would also move the invariant comment away from the code it constrains.

The underlying concern — a future subscope relocation silently journaled as `Intra` — is real, but the helper does not address it. A constant-returning function fails open on that day identically to a literal; only real classification logic fixes it, and this session cannot write it because it resolves exactly one scope and has no view of a destination outside it. Meanwhile the publish path is fail-closed: `publish_op` matches `crossing: ScopeCrossing::Intra` explicitly and sends every other crossing to `Err(Halt::Unclassified)` rather than publishing a plain ref move. The subscope slice will replace two literals with real derivation; a placeholder now buys nothing and costs a review cycle.

### 3. `sync/drain.rs` — record why `Halt::Unclassified` is right here — **declined**

Two of the three proposed sentences are shapes `AGENTS.md` bans outright. "The facade authors only `Intra` today, so nothing reaches here" is an absence-justifying comment — prose defending a path that is not in the code, which becomes a falsehood in the same diff that wires the path. "The subscope slice must land the re-seal plan before it can" states a plan, not a property of the code.

The third sentence is genuine rationale, and it is already stated at its home — twice. `Halt::Unclassified`'s own doc comment carries it at the definition ("Charged nothing and retried on the next tick, so an outage never abandons an op"), and `publish_ref_move` restates the consequence where a caller must branch on it ("the one verdict that retries free and forever"). `AGENTS.md` asks for non-obvious rationale **once, at its home, not restated on every caller**. A third copy at this match arm is the restatement the rule prohibits. The comment already there explains the non-obvious part — why publishing as a plain ref move would be wrong — which is the piece not derivable from the `Halt` definition.

### 4. `rotation/trigger.rs` — plan durable retention for failed triggers — **declined**

This asks for a comment documenting a requirement for a caller that does not exist, about a store that was deliberately not built. Both halves are banned shapes: absence-justification for the unlanded integration, and a plan for future work with no code to check it against — the same rot that `pnpm lint:tracker-refs` gates against, just spelled in prose instead of `#NNNN`.

The retention law is also already stated at its home. `ScopeExitReport::failed` documents it on the field ("Each is still a live trigger: the caller re-drives it rather than treating the pass as done") and `is_complete()` documents the contract on the method ("the only state in which the caller may consider the scope exits settled"). The design rationale for having no second store is in the PR body: the halted op stays at the queue head, so every pass re-derives the trigger from the queue, and a durable trigger store would duplicate state the queue already holds. The premise that a lost trigger strands a departed grantee does not hold while the op is not dequeued.

---

Gates on `2a32661d5`, all exit 0: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets`, `cargo test -p cipherbox-engine -p cipherbox-core`, `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown`, `node scripts/check-tracker-refs.mjs`.

Issue #872, opened 2026-07-28
engine: prune old versions and reclaim their bytes


FSM1:
Prune drops old file versions and reclaims their bytes. Prune is a **write-plane mutation** — it re-seals and publishes a shortened history — which is why #820 pulled it into this chain. Slice settled by #825 on wayfinder map #813.

## Scope

**This issue absorbs #849, #850 and #851 outright.** They were previously named as prerequisites in prose only, with no dependency edge, which made #872 look startable when it is not — and made three defects look like three shippable PRs. They are not: `plan_prune` (`crates/engine/src/content/retention.rs:93`) has **zero production callers** at `main` @ `3d087d673`. Its only references are the re-exports at `crates/engine/src/lib.rs:54` and `crates/engine/src/content/mod.rs:38`, and its own unit tests in `retention.rs`. Nothing reaches it, so none of the three is independently observable or independently testable end-to-end. Close all three as folded-in when this ships. Native `blocked_by` edges are recorded on this issue so the graph stays honest until they are.

**Three correctness fixes, folded in.**

- **#849 — leaf expansion.** `plan_prune` emits root content CIDs only, and `POST /registry/retire` deletes rows for exactly the CIDs handed to it without walking the DAG, so every sealed leaf stays pinned and charged forever while `reclaimed_bytes` reports the whole version freed. Expand each doomed root to `root + leaf_cids` before retiring, by fetching the root block and `decode_root`-ing it — the root block is plaintext det-CBOR, so no key is needed. Two constraints on its shape: `root + leaf_cids` **cannot be one call** — the registry batch cap is `REGISTRY_BATCH_MAX = 1000` (`crates/engine/src/net/mod.rs:25`), enforced by the chunking at `crates/engine/src/net/retire.rs:100`, and a 1 GiB version is 1025 targets at the frozen 1 MiB framing, so partial retire is normal at exactly `max_file_bytes` — and **leaves retire before their root**.
- **#850 — unit mismatch.** `ContentVersion.size_bytes` (`crates/engine/src/content/retention.rs:75`) is documented as the **pinned** size, but the prune path fills it from `Version.size`, which is **plaintext** bytes. The gap is fixed and computable (40 B per sealed leaf, plus the root block). A pending-reclaim total built on the wrong unit is the same misreport in a new place. Derive the pinned total from the leaf count and root size rather than reusing the plaintext number.
[...]

PR #1115, opened 2026-08-06
fix: expand pruned roots to their leaves and plan reclaim in pinned bytes

### LINKED ISSUE #849: engine: prune retires only the DAG root, orphaning every sealed leaf
## Problem

`plan_prune` emits **root content CIDs only**, but `POST /registry/retire` deletes rows for exactly the CIDs handed to it and does not walk the DAG.

- `PrunePlan.retire_targets` is built from `ContentVersion.content_cid` — the version's DAG root (`crates/engine/src/content/retention.rs:47-55,80`).
- The API side does `pinRepo.delete({ accountId, cid: In(targets) })` and then unpins only CIDs that reach global-zero refcount (`apps/api/src/registry/services/registry.service.ts`, `retire`). It cannot expand a root: it is zero-knowledge and never interprets content.

So retiring a version deletes the root's pin row and **leaves every sealed leaf pinned and charged to the account forever**, while `reclaimed_bytes` (`retention.rs:81`) reports the whole version as freed. Prune therefore reclaims ~0% of a multi-chunk version's bytes and reports 100%.

At the frozen 1 MiB framing this is the normal case, not an edge case: any file over one chunk has more leaf bytes than root bytes.

## Fix

Expand each doomed root to its full CID set before retiring: fetch the root block, `decode_root` it (`crates/engine/src/content/dag.rs:159-199`), and retire `root + leaf_cids`.

No key is required and no wire-format change is needed — the DAG root block is **plaintext** deterministic CBOR (`assemble` performs no seal, `dag.rs:124`), so the leaf list is readable by CID alone. Cost is one block fetch per pruned version.

Rejected alternative: persisting the leaf CID set in `Version` (`crates/core/src/seal/body.rs:140-152`). That is a wire-format addition to the struct frozen by #820, to store data the root block already carries authoritatively.

Also fix `reclaimed_bytes` to sum what is actually retired — see the sibling unit-mismatch defect.

## Ordering

Publish the shortened version history **before** retiring. Retire-first leaves a reader resolving the still-live old record against CIDs that have been unpinned.

Found while resolving #820.

Closed by slice #872.

Part of #655

### LINKED ISSUE #850: engine: ContentVersion.size_bytes conflates plaintext size with pinned size
## Problem

Two fields that flow into each other measure different things:

- `Version.size` is documented as **plaintext** bytes — "The plaintext size in bytes" (`crates/core/src/seal/body.rs:144-145`).
- `ContentVersion.size_bytes` is documented as the **pinned** size — "The version's pinned size in bytes (what a prune reclaims)" (`crates/engine/src/content/retention.rs:53-54`).

The prune path reads the version history out of `ReadBody::File { versions }` (`body.rs:246`) and maps it into `ContentVersion`, so the plaintext number lands in a field that means pinned bytes. `plan_prune` then sums it into `reclaimed_bytes` (`retention.rs:81`) and reports it to the user.

## The gap

Sealed bytes exceed plaintext by a fixed, computable amount:

- 40 B per leaf — a 24-byte XChaCha20-Poly1305 nonce plus a 16-byte tag (`crates/engine/src/content/chunk.rs:65,93`)
- plus the DAG root block itself: 41 B fixed + 38 B per inlined leaf CID

At the #820-frozen framing (`chunk_size = 1_048_536`, sealed leaf exactly 1 MiB) that is 78 B per chunk, so the under-report is small in ratio but always present and always in the same direction: **`reclaimed_bytes` understates what is actually freed**, on top of the much larger overstatement from the root-only retire defect. The two errors are independent and point opposite ways, so neither masks the other reliably.

~~The same conflation reaches the quota pre-flight.~~ **Stale — corrected 2026-08-07 against `main` @ `2819cb831`.** The pre-flight is already sized in sealed bytes: `crates/engine/src/facade.rs:2370` computes `requested = sealed_total_bytes(size, profile)` and passes it to `pre_flight_quota_check` at `facade.rs:2457`, which is what #822 §5 required. This issue's live surface is only `ContentVersion.size_bytes` and the `reclaimed_bytes` it feeds — and that surface has no producer at all, since nothing maps `ReadBody::File { versions }` into `ContentVersion` outside a test helper.

## Fix

Pick one unit per field and make the conversion explicit at the boundary rather than implicit in an assignment. Derive the pinned total from the leaf count and root size rather than reusing the plaintext number, and add a test asserting `reclaimed_bytes` equals the sum of the sizes the retire call actually frees.

Found while resolving #820.

Closed by slice #872.

Part of #655

### LINKED ISSUE #851: engine: plan_prune keep_latest admits zero, retiring the live version
## Problem

`plan_prune(versions_newest_first: &[ContentVersion], keep_latest: usize)` (`crates/engine/src/content/retention.rs:72`) accepts `keep_latest = 0`, which places every version — including the live one — into `retire_targets`.

There is no guard: `versions_newest_first.iter().skip(0)` yields the whole history (`retention.rs:73`). The result is a plan that retires the file's current content while the file record still points at it.

## Fix

Make it unrepresentable rather than checked: take `NonZeroUsize`. A prune that keeps nothing is not a valid retention policy, and the existing tests at `retention.rs:141-163` only cover `keep_latest >= 1`, so nothing depends on the zero case.

This is the same move already made one module over for the chunk size: `ContentProfile::new` rejects zero at construction so a zero chunk size is unrepresentable rather than a fail-late panic (`crates/engine/src/content/profile.rs:48-54`). Consistent with that precedent, and with the codebase's fail-closed posture.

Small and self-contained — roughly a three-line change plus a signature update at the call sites.

Found while resolving #820.

Closed by slice #872.

Part of #655

FSM1:
Three correctness defects on the prune path, all in `crates/engine/src/content/retention.rs`.

## \#849 — a root-only retire orphans every sealed leaf

`plan_prune` emitted root content CIDs only, but `POST /registry/retire` deletes rows for exactly the CIDs handed to it and never interprets content — it is zero-knowledge and cannot expand a root. Retiring a version therefore dropped the root's pin row and left every sealed leaf pinned and charged forever, while `reclaimed_bytes` reported the whole version freed. At the frozen 1 MiB framing any file over one chunk has more leaf bytes than root bytes, so this was the normal case.

New `expand_retire_targets(version, root_block)`. The root block is plaintext deterministic CBOR, so the leaf list is readable by CID alone with no key and no wire-format change.

- **Fail-closed throughout.** The block is refused past `MAX_RESOLVED_RECORD_BYTES` before it is hashed, then verified against the version's `contentCid` through `decode_content_cid_str` + `is_plane_anchor` + `verify_cid` — the same spelling `read.rs` and `staging.rs` already use, so the trust failure carries core's `TrustViolation::ContentCidMismatch` and an off-plane root CID is refused as malformed rather than conflated with a substituted block.
- **The walk is one level.** `decode_root` already enforces that every link is a `raw`-codec CID, and a raw block names no further links.
- **Leaves first, root last.** The root block is the only record of its own leaf list, so retiring it first would make those leaves unnameable and charged forever — exactly the defect, made permanent instead of transient. A drain that dies mid-expansion can always re-expand from a root that is still pinned.

`plan_prune` still emits roots only; the expansion is the drain-time step, matching the ledger-stores-roots decision in \#872. The API needed no change: the existing `retire` already takes an advisory lock on every path, bulk-deletes via `In(targets)`, and refcounts before unpinning, which is exactly right once the client hands it the full set.

## \#850 — plaintext bytes in a field meaning pinned bytes

`ContentVersion.size_bytes` was documented as the pinned size but was fed from `Version.size`, which is plaintext. Sealing adds 40 B per leaf plus the root block, and pinned bytes are what the registry charges, so the figure understated every reclaim and every quota pre-flight.

The field is now `pinned_bytes`, and `ContentVersion::from_plaintext_size(cid, plaintext_bytes, profile)` is the conversion — explicit at the boundary rather than implicit in an assignment, and it takes the framing profile because plaintext cannot be converted without it. It reuses the existing `sealed_total_bytes`, which derives the total from the leaf count and root-block length and is already held to real staging byte-for-byte.

The two figures are then bound **at runtime, not just in a test**: `expand_retire_targets` refuses unless the manifest's own leaf count and actual root-block length account for exactly the pinned total the plan quoted. That is \#872's acceptance criterion "`reclaimed_bytes` equals the sum of the sizes the retire call actually frees, in pinned bytes", enforced rather than asserted — and it closes the manifest's freedom to declare an arbitrary `size`/`chunkSize` pair, since those now have to reproduce the planned figure.

## \#851 — `keep_latest` admitted zero

`plan_prune` took a plain `usize`, so `skip(0)` yielded the whole history including the live version the file record still points at. It now takes `NonZeroU64`, making the state unrepresentable rather than guarded, and matching the `RetentionPolicy::KeepLatest(NonZeroU64)` the call site already holds — no lossy conversion at the boundary. The enum was already correct and is untouched. A keep-count past `usize` on a 32-bit target (wasm32 is a shipped target) clamps to keeping everything.

## Gate

`Engine Tests`. Fifteen unit tests in `crates/engine/src/content/retention.rs`, all behavioral. The fail-closed checks return `Err` rather than asserting, and the suite is verified green under `cargo test --release` so no check is compiled out.

`a_gibibyte_version_expands_past_one_retire_batch_with_the_root_last` pins the shape constraint: at the frozen framing a 1 GiB version expands to 1025 targets, past `REGISTRY_BATCH_MAX = 1000`, and the root still lands in the final batch after every leaf it names.

## Review gates

`/simplify`, `/security-review` and `/crypto-privacy-review` all ran on `git diff main...HEAD`. Folded in: routing the address check through the existing core primitives instead of a bespoke multibase string compare; binding the drain-time pinned total to the plan (which collapsed a `RetireExpansion` struct into a plain target list); removing an invariant restated on four call sites and an absence-justifying comment about recursion the function does not do; deleting a doc claim that a constructor was the only way to build a struct with public fields; a redundant test folded into its neighbour; and `with_capacity` instead of a `collect`-then-`push` realloc.

Two findings were deliberately **not** applied literally and are filed instead, both blocked on this issue's parent:

- \#1124 — the expansion cannot subtract a leaf that a *retained* version also references, so an aliased root could unpin live content. Not reachable for honestly-authored versions (fresh per-version key and per-chunk nonce make leaf sets disjoint), and the retained set is a whole-plan property only the prune op knows. Passing a retained set into a single-root expansion ahead of that caller would be speculative and would break the closed-form byte accounting.
- \#1125 — `registry_cids` in `sync/drain.rs` derives the same CID set for the register/abandon path, with the opposite ordering. Both orderings are correct for their own path because the expansion key differs (staging there, the root block here), but nothing holds the two *sets* equal. Unifying them wants the prune drain in place, and `sync/` was owned by a parallel branch this wave.

## Scope

\#872 is **not** closed by this. These three were its sub-issue tree and its remaining open blockers, but its own acceptance goes further: the prune op that ends at publish, the `RetireLedger` seam and its conformance kit, the resumable drain, the `retired: 0` done-signal, and the vault-level pending-reclaim figure with its mode-derived legs. This PR lands the three prerequisites \#872 folded in, in the shapes \#872 specified, and leaves the ledger to \#872 itself.

No TypeScript changed, so the TS gates are unaffected.

Closes #849
Closes #850
Closes #851
Part of #872

### DIFF
--- a/crates/engine/src/content/mod.rs
+++ b/crates/engine/src/content/mod.rs
@@ -35,7 +35,8 @@ pub use read::{
     leaf_range_for_byte_range, read_block,
 };
 pub use retention::{
-    ContentVersion, PrunePlan, QuotaExceeded, RetentionPolicy, plan_prune, pre_flight_quota_check,
+    ContentVersion, ExpandError, PrunePlan, QuotaExceeded, RetentionPolicy, RetireExpansion,
+    expand_retire_targets, plan_prune, pre_flight_quota_check,
 };
 pub use write::{ContentWriter, FinishedContent};
 
--- a/crates/engine/src/content/mod.rs
+++ b/crates/engine/src/content/mod.rs
@@ -35,8 +35,8 @@ pub use read::{
     leaf_range_for_byte_range, read_block,
 };
 pub use retention::{
-    ContentVersion, ExpandError, PrunePlan, QuotaExceeded, RetentionPolicy, RetireExpansion,
-    expand_retire_targets, plan_prune, pre_flight_quota_check,
+    ContentVersion, ExpandError, PrunePlan, QuotaExceeded, RetentionPolicy, expand_retire_targets,
+    plan_prune, pre_flight_quota_check,
 };
 pub use write::{ContentWriter, FinishedContent};
 
--- a/crates/engine/src/content/retention.rs
+++ b/crates/engine/src/content/retention.rs
@@ -223,7 +223,9 @@ mod tests {
     use super::*;
     use crate::content::chunk::{ContentKey, frame_and_seal};
     use crate::content::dag::assemble;
+    use crate::net::REGISTRY_BATCH_MAX;
     use crate::testkit::SeededEntropy;
+    use cipherbox_core::content::{CONTENT_CID_CODEC, compute_cid};
     use cipherbox_core::suite::aead::KEY_LEN;
 
     fn hosted(used: u64, limit: u64) -> Quota {
@@ -381,6 +383,35 @@ mod tests {
             .expect("at least the root");
         assert_eq!(leaves, leaf_cids, "every leaf, in file order");
         assert_eq!(root, &root_cid, "the expansion key retires last");
+    }
+
+    /// At the frozen 1 MiB framing a 1 GiB version expands past the registry's
+    /// batch cap, so a prune spanning several batches is the normal case rather
+    /// than an edge one — and the root must still land in the last batch.
+    #[test]
+    fn a_gibibyte_version_expands_past_one_retire_batch_with_the_root_last() {
+        let profile = ContentProfile::PRODUCTION;
+        let leaves = 1024usize;
+        // Address-only: a 1 GiB version's leaf *blocks* are irrelevant to an
+        // expansion, which reads the root's link list alone.
+        let leaf_cids: Vec<Vec<u8>> = (0..leaves)
+            .map(|i| compute_cid(CONTENT_CID_CODEC, &(i as u64).to_be_bytes()))
+            .collect();
+        let plaintext_len = leaves as u64 * profile.chunk_size() as u64;
+        let dag = assemble(&leaf_cids, plaintext_len, &profile).expect("assembles");
+        let root_cid = encode_content_cid_str(&dag.content_cid);
+
+        let expansion = expand_retire_targets(&root_cid, &dag.root_block).expect("expands");
+        assert_eq!(expansion.targets.len(), leaves + 1);
+        assert!(
+            expansion.targets.len() > REGISTRY_BATCH_MAX,
+            "the expansion cannot ride one retire call"
+        );
+        assert_eq!(
+            expansion.targets.last().expect("non-empty"),
+            &root_cid,
+            "the root rides the final batch, after every leaf it names"
+        );
     }
 
     /// The plan-time figure and the drain-time figure are two derivations of one
--- a/crates/engine/src/content/retention.rs
+++ b/crates/engine/src/content/retention.rs
@@ -178,10 +178,9 @@ impl From<DagError> for ExpandError {
 /// drain that dies mid-expansion can re-expand from a root that is still pinned,
 /// where a root-first order would leave its leaves unnameable and charged.
 ///
-/// Fail-closed throughout. The block is refused past the resolved-block ceiling
-/// before it is hashed, verified against `version.content_cid`, decoded, and
-/// finally held to the pinned total the plan quoted — a version whose bytes
-/// cannot be accounted for is never retired.
+/// The manifest is held to the pinned total the plan quoted, so a prune can never
+/// free bytes it did not report — and a `size`/`chunkSize` pair the link count
+/// alone cannot constrain has to reproduce that figure.
 pub fn expand_retire_targets(
     version: &ContentVersion,
     root_block: &[u8],
--- a/crates/engine/src/content/retention.rs
+++ b/crates/engine/src/content/retention.rs
@@ -8,6 +8,13 @@
 
 use core::num::NonZeroU64;
 
+use cipherbox_core::content::encode_content_cid_str;
+
+use super::budget::sealed_total_bytes;
+use super::chunk::SEALED_LEAF_OVERHEAD;
+use super::dag::{DagError, decode_root, root_block_cid};
+use super::limits::MAX_RESOLVED_RECORD_BYTES;
+use super::profile::ContentProfile;
 use crate::api::Quota;
 
 /// A pre-flight quota rejection: the hosted account cannot admit `needed_bytes`.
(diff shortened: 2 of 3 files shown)


FSM1:
Closing unmerged: #849, #850 and #851 are folded into #872 rather than shipped separately.

`plan_prune` has zero production callers, so nothing here is reachable or observable at runtime, and #872 would immediately reshape the expansion around the retire ledger. #872's own body already absorbs all three and carries `Closes #849, #850, #851`, so they close when it ships. That was the recommendation in the audit comments on all three issues; shipping this PR would have landed an expansion with no caller.

**The branch is kept, not deleted** — `fix/prune-leaf-expansion-and-pinned-units` @ `head`. It is a working reference for #872's implementer, and lands in the shapes #872 specifies:

- `expand_retire_targets(version, root_block)` returns leaves first and the root **last**, one level, block-size-capped before hashing. The address check routes through core's `decode_content_cid_str` + `is_plane_anchor` + `verify_cid`, so a mismatch carries the already-KAT-pinned `TrustViolation::ContentCidMismatch`.
- `size_bytes` -> `pinned_bytes` with `ContentVersion::from_plaintext_size(cid, plaintext, profile)` as the explicit conversion, bound at runtime: the expansion refuses unless the manifest's leaf count and root-block length account for exactly the pinned total the plan quoted.
- `plan_prune` takes `NonZeroU64`, matching the `RetentionPolicy::KeepLatest(NonZeroU64)` the call site already holds.

Verified along the way and worth keeping: `registry.service.ts` `retire` needs **no change** — it already advisory-locks every path, bulk-deletes via `In(targets)`, and refcounts before unpinning. It was only ever missing the full target set from the client.

Two findings from this work are filed against #872 and stay open: #1124 (the expansion cannot subtract a leaf a retained version also references — adversarially reachable via a write-grantee) and #1125 (`sync/drain.rs`'s `registry_cids` derives the same CID set with the opposite ordering, and nothing holds the two equal).

Issue #1019, opened 2026-08-03
engine: add the write-revoke and downgrade rotation trigger


FSM1:
## Already landed — do not re-implement

`rotate_scope_write` (`crates/engine/src/rotation/rotate_write.rs`) — the commitment re-sign, the fresh write override seed, the child-first name wave with the root re-pointed last, `derive_write_name`, `build_repoint_object`, the three-channel re-point, register-first inventory swap and root linger — is written and simulation-tested. `revoke_read_grant` and the eager cascade are landed in `crates/engine/src/rotation/trigger.rs` and `cascade.rs`. **The write-rotation primitive is not in scope here.**

## Scope

`RotationTrigger` at `crates/engine/src/rotation/trigger.rs` enumerates only three of the blueprint's five triggers — `ScopeExit`, `ReadRevoke`, `Manual`. This adds the fourth: **write revoke / downgrade**.

- extend `RotationTrigger` with the write-revoke/downgrade variant and its host-facing `name()` string, and extend the module's trigger table doc to match the blueprint's
- the owner-only committed-set cut for the write plane, mirroring what `revoke_read_grant` does for the read plane: remove or demote the recipient in the grant ledger and the grant-set commitment, owner-re-sign, then drive `rotate_scope_write`
- **full revoke additionally rotates the read plane** — write revoke is `rotateScopeWrite` *plus* a read rotation; a downgrade to read is the write rotation alone, and the downgraded recipient keeps a live read grant
- flag the residual honestly to the host: old write names are hijackable by the revokee and are therefore dead to survivors — tombstones are advisory only, and the forgery window stays wave-bounded for write-grantee survivors

## Blueprint anchors

`blueprint/engine.md`: Rotation primitives — Triggers ("Write revoke / downgrade → `rotateScopeWrite`; plus read rotation on full revoke") and Residuals; Grants and ledger — Revocation is discovered, not delivered ("Write revoke/downgrade = write rotation; old names are hijackable by the revokee and therefore dead to survivors").

`CONTEXT.md`: name wave, forgery window, write-scope cut, grant ledger, grant-set commitment, root linger.

## Gate

Engine simulation:

- a downgrade rotates the write plane only; the recipient's read grant is still live and their blob still present at their tag afterwards
- a full write revoke rotates both planes; the revokee has no blob at their tag on the new read-plane record
- the revoked writer is absent from the re-signed commitment and the ledger
- surviving write-grantees derive every new name locally — zero re-discovery
[...]

PR #1116, opened 2026-08-06
feat: implement the concrete WriteSubtreeResolver over the real seams

### LINKED ISSUE #1108: engine: implement the concrete WriteSubtreeResolver over the real seams
## Already landed — do not re-implement

`rotate_scope_write` and the whole write-plane wave are written and simulation-tested in `crates/engine/src/rotation/rotate_write.rs`: the commitment re-sign, the fresh write override seed, the child-first name wave with the root re-pointed last, `derive_write_name`, `build_repoint_object`, the three-channel re-point, register-first inventory swap and root linger.

The wave's **write edge** is production code: `WriteWavePublisher for WriteWaveNet` at `crates/engine/src/net/rotation.rs:1248` (#1101). **That is not in scope here.**

## Scope

`WriteSubtreeResolver` is the wave's **read edge** — the analogue of the cascade's `CascadeResealResolver` — and it has **no production implementation**. Every implementor is a test fake: `FakeResolver`, `PostFlipResolver` and `FailingResolver` at `crates/engine/src/rotation/rotate_write.rs:761`, `:784`, `:805`, all inside the test module opening at `:698`. The trait's own doc says so (`rotate_write.rs:150`): *"the real network/gate wiring is not landed and tests fake it."*

The trait is one method (`rotate_write.rs:154`):

```rust
async fn resolve_node(&self, node_id: &[u8; 16]) -> Result<WriteScopeNode, ResolveFailure>;
```

returning `node_id`, `current_name` and `child_node_ids` (`rotate_write.rs:77`). Land it over the real seams:

- **resolve + adoption-gate + unseal behind the trait.** Reuse `OwnerRotationNet`'s gated read (`crates/engine/src/net/rotation.rs`) rather than growing a parallel one — every record the enumeration consumes must have passed the adoption gate, and a gate failure is a fail-closed trust verdict, never staleness (rule 6).
- **`ResolveFailure` must keep the trust/availability split intact.** The enumeration at `rotate_write.rs:671-682` walks the subtree breadth-first through this seam; a record that cannot be authoritatively obtained must not be laundered into a retryable stall, and a node the walk cannot resolve must abort the wave rather than silently truncate the subtree. A truncated enumeration re-points the root while leaving interior nodes un-rotated — the revokee keeps a live write name under a root that claims to have moved.
- **child-rewrite coverage.** `WriteSubtreeResolver::resolve_node`'s doc carries the coverage obligation the production impl must satisfy; #1101's body defers it here explicitly.

## Why this is filed separately

It was **never in #1014's scope** — that issue named four seams and this is not one of them — so nothing mis-closed. #1027 (which closed #1014) deferred `SweepResolver` to #1025 and `WriteWavePublisher` to #1026, and both were honoured. `WriteSubtreeResolver` simply had no successor issue. #1101's body says its production implementation "is a separate issue"; this is that issue.

Note `rotate_scope_write` consequently has **no production caller** on `main` — outside its own module it is referenced only by a doc comment (`crates/engine/src/grants/create.rs:21-22`) and the `rotation/mod.rs` re-export (`:58-59`).

## Blueprint anchors

`blueprint/engine.md`: Rotation primitives — `rotateScopeWrite`, the child-first name wave, the three-channel re-point; Pointer planes — register-first inventory swap and root linger.

`CONTEXT.md`: name wave, write plane, register-first, root linger, adoption gate, forgery window.

## Gate

Engine integration suite against the fake network world, not inline hand-rolled fakes:

- every record the enumeration consumes has passed the adoption gate; a gate rejection aborts the wave and is never reported as staleness
- a node that cannot be authoritatively resolved aborts the wave rather than truncating the subtree — assert the root is **not** re-pointed in that case
- the production resolver and the inline fakes agree on the same simulation scenarios that `rotate_scope_write` is already tested against
- a resumed wave re-resolves from published records only, with no in-memory carry (the existing `is_republished` contract, now over the real read edge)

## Depends on

Nothing outstanding. The primitive, and the wave's write edge (`WriteWavePublisher`), are landed.

## Blocks

- **#1019** — write revoke / downgrade drives `rotate_scope_write`, which cannot run without this read edge. #1019's `Depends on` names only a production `WriteWavePublisher`; that is now landed, so without this edge recorded #1019 would read as unblocked while still being blocked.
- **#764** — reclaiming orphaned interior write-names after a post-flip crash operates on the same enumeration.

Part of #635

FSM1:
## What

`WriteSubtreeResolver` — the write-plane name wave's **read edge** — had no production implementation. Every implementor (`FakeResolver`, `PostFlipResolver`, `FailingResolver`) lived inside `rotate_write.rs`'s `#[cfg(test)]` module, so `rotate_scope_write` compiled green and could not run in production. This lands the concrete resolver over the real seams.

`WriteWaveNet` now implements it alongside the `WriteWavePublisher` it already carried, so one value serves both edges of the wave — the shape `OwnerRotationNet` already uses for `ChildIndexResolver` / `CascadeResealResolver` / `ScopeRootPublisher`.

## How

- **Resolve + adoption-gate + unseal reuse the wave's own reads.** `resolve_node` runs `root_source` for the scope root and `interior_source` for everything below it, so every record the enumeration consumes passes the adoption gate under the caller's own scope/node label — no parallel gated read was grown.
- **Names come only from a gated parent.** `WaveSubtree` records each child's name as the descent gates its parent, so no caller-supplied label ever chooses where a node is read, and an id nothing gated named is a fail-closed refusal rather than a fetch. One id named at two **different** names aborts with `ConflictingChildLabel` — picking either would rotate one name and leave the other live. A benign `#33 D5` dual link names the child identically in both parents and is unaffected.
- **`child_node_ids` stops at a *proven* descendant scope root** — see the security section below.
- **Verdicts keep rule 6's split in both directions.** `subtree_verdict` carries only a fail-closed refusal across as `ResolveFailure::Rejected`; availability stays availability. An unresolvable node aborts the wave before the root is re-pointed, so a truncated enumeration can never leave a revokee holding a live write name under a root that claims to have moved.

### Two defects fixed along the way

**1. A resumed wave could never re-read its own root.** `root_source` now falls back to the at-floor recovery (`reread_at_floor` — the path `OwnerRotationNet::gated_root` already takes) when the root's record comes back at exactly the sequence floor its previous adopt raised. Without it the second `adopt_root` rejects as not-newer and the rotation is permanently stuck on a `Rejected` verdict no retry clears. `a_resumed_wave_re_resolves_the_subtree_from_published_records_alone` fails with `Rejected` if the fallback is removed.

**2. An unverified `directChildScopeIndex` carved nodes out of the revocation.** The `/security-review` and `/crypto-privacy-review` gates independently found the same fail-open in the first draft of this PR. That field is sealed under the root's `writeKey` and authenticated by any **committed write pseudonym** (`committed_write_pseudonyms` admits every `Permission::Write` entry, not just the owner) — so the grantee a write revoke is cutting can author it. An entry naming an ordinary in-scope node made the wave skip that node: never republished, its `ChildRef.ipnsName` left untouched by the parent's rewrite, its old name never retired. The revokee kept a live write name under a root claiming to have moved — permanently, not for the wave-bounded forgery window the residuals accept.

Only the owner-signed grant-set commitment names a scope root's own `ipnsName`, so `record_scope_boundary` now gates each claimed entry through the root gate and refuses a boundary it cannot prove. The read plane already gates this same index; the write plane used it to *exclude*, which is the direction that had to fail closed.

## Tests

New, in `net/rotation.rs` against the fake network world (the existing `Test` gate — no new suite):

- `the_enumeration_descends_gated_records_and_stops_at_a_descendant_scope_root`
- `a_planted_index_entry_never_exempts_an_ordinary_node_from_the_wave` — the fix above
- `a_node_no_gated_record_named_has_no_name_to_read_at`
- `one_id_named_at_two_names_aborts_rather_than_picking`
- `a_gate_rejected_node_aborts_the_wave_and_is_never_reported_as_staleness` — asserts the scope pointer and the new root name carry **nothing**
- `an_unresolvable_node_aborts_the_wave_rather_than_truncating_the_subtree`
- `the_production_edges_carry_the_whole_wave_from_enumeration_to_re_point` — full `rotate_scope_write` over the production resolver *and* publisher, then a read-only holder descends root → mid → leaf by rewritten child refs alone while the descendant scope root keeps its own name
- `a_resumed_wave_re_resolves_the_subtree_from_published_records_alone`

## Review gates

All three ran on this diff. Beyond the HIGH above, `/simplify` produced the `gated_scope_root` extraction (the adopt/at-floor ladder was written three times once this branch added the third) and corrected two doc blocks this branch falsified — `GatedWaveRoot`'s rationale and a test comment both claimed the root gate has no at-floor re-open, which is precisely what this branch adds.

Two suggestions were verified wrong and rejected: replacing the `read_scope_seed` unwrap with `unwrap_or_default()` (would derive a read key from an all-zero seed) and flattening `RefCell<Discovered>` into two cells (`RotationAncestry` in the same file establishes that convention).

## Gates

| Gate | Exit |
| --- | --- |
| `cargo fmt --all --check` | 0 |

No TypeScript surface is touched.

## Out of scope

Two review findings land in `#1101`'s code rather than this read edge, and both are filed blocking #1019 — the production caller — so the write revoke cannot ship over them:

- #1129 — the re-seal mints grant blobs from the record's committed set, not the owner-verified plan commitment
- #1130 — the root republish checks the write-epoch floor against an enumeration-time snapshot

`rotate_scope_write` still has no production caller; that is #1019.

Closes #1108

> [!NOTE]
> ### Implement production `WriteSubtreeResolver` over real seams in write-wave enumeration
> - Implements `WriteSubtreeResolver::resolve_node` on `WriteWaveNet` to enumerate nodes from published records, halting at proven descendant scope roots and failing on conflicting child labels.
> - Adds `WaveSubtree` to track discovered node names and proven descendant scope boundaries per enumeration pass, with conflict detection via `record_children`.
> - Adds `WriteWaveNet::record_scope_boundary` to validate claimed descendant scope roots by gating each against the parent-derived seed before stopping enumeration.
> - Extracts `gated_scope_root` helper to commonize root adoption/gating logic and normalize `GateError` to `ResolveFailure` across callers.
> - Adds `wave_read_verdict` and `subtree_verdict` utilities to consistently translate between `ResolveFailure` and `WritePublishError` on read and enumeration paths.
> - Risk: `read_epoch` is now taken from `envelope.epoch` rather than `outcome.adopted.epoch`; absent `read_scope_seed` now yields `NotLanded` via earlier gating rather than a separate code path.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 42018af.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Improved write rotation to discover complete in-scope trees while respecting authenticated descendant boundaries.
  * Added validation for scope boundaries and conflicting labels during rotation.
  * Rotation now resolves nodes from published records before republishing.

* **Bug Fixes**
  * Improved handling of unavailable or rejected records and resumed rotation walks.
  * Standardized error reporting across rotation and subtree-resolution steps.

* **Documentation**
  * Updated rotation documentation to reflect live subtree resolution and publishing support.

### DIFF
--- a/crates/engine/src/net/rotation.rs
+++ b/crates/engine/src/net/rotation.rs
@@ -13,7 +13,8 @@
 //! pipeline stays the only path to the transport.
 
 use core::cell::RefCell;
-use std::collections::BTreeMap;
+use std::collections::btree_map::Entry;
+use std::collections::{BTreeMap, BTreeSet};
 
 use cipherbox_core::ipns::IpnsName;
 use cipherbox_core::kdf;
@@ -50,7 +51,7 @@ use crate::rotation::{
     CascadeResealResolver, CascadeTarget, ChildIndexResolver, CommittedSet, RepointChannel,
     RepublishedNode, ResealError, ResealSeeds, ResealedScopeRoot, ResolveFailure,
     ScopeRootIdentity, ScopeRootPublishError, ScopeRootPublisher, WritePublishError,
-    WriteWavePublisher, reseal_scope_root,
+    WriteScopeNode, WriteSubtreeResolver, WriteWavePublisher, reseal_scope_root,
 };
 use crate::seams::{CredentialStore, FloorStore, Http, RecordTransport, Scheduler};
 use crate::session::SessionIdentity;
@@ -688,8 +689,9 @@ where
 }
 
 /// The write-plane name wave's transport edge (blueprint/engine.md
-/// "rotateScopeWrite"): the owner arm of [`WriteWavePublisher`] over the same net
-/// plane the read-plane seams above run on.
+/// "rotateScopeWrite"): the owner arm of both [`WriteSubtreeResolver`] and
+/// [`WriteWavePublisher`] over the same net plane the read-plane seams above run
+/// on.
 ///
 /// The wave hands this no authoring material — only routing identity and the one
 /// write seed that signs at the new name ([`RepublishedNode`]) — so every
@@ -742,6 +744,72 @@ pub struct WriteWaveNet<'a, T, H: Http, C: CredentialStore, F, Sch, E> {
     /// The root read this pass gated and has not yet republished
     /// ([`GatedWaveRoot`]). One rotation pass per net.
     pub gated_root: GatedWaveRoot,
+    /// The subtree index the enumeration builds as it descends
+    /// ([`WaveSubtree`]). One rotation pass per net.
+    pub subtree: WaveSubtree,
+}
+
+/// The write scope's node index, as this pass's own gated reads discovered it.
+///
+/// A node id locates nothing on its own — only a gated parent's read body names
+/// its children — so the enumeration's descent is what supplies every name below
+/// the root, and no caller-supplied label ever chooses where a node is read.
+#[derive(Default)]
+pub struct WaveSubtree {
+    inner: RefCell<Discovered>,
+}
+
+#[derive(Default)]
+struct Discovered {
+    names: BTreeMap<[u8; 16], IpnsName>,
+    /// The root's directly-descendant scope roots: each rotates under its own
+    /// write scope seed, so the wave stops at them (`grants/child_index.rs`).
+    child_scopes: BTreeSet<[u8; 16]>,
+}
+
+impl WaveSubtree {
+    /// The name a gated parent gave `node_id`.
+    fn name(&self, node_id: &[u8; 16]) -> Option<IpnsName> {
+        self.inner.borrow().names.get(node_id).cloned()
+    }
+
+    fn record_child_scopes(&self, index: &[ChildScopeRef]) {
+        self.inner
+            .borrow_mut()
+            .child_scopes
+            .extend(index.iter().map(|child| child.scope_id));
+    }
+
+    /// Record `body`'s in-scope children at the names it gives them and return
+    /// their ids for the walk to descend into.
+    ///
+    /// Two parents naming one id at **different** names is the read plane's C2
+    /// conflict ([`ResolveFailure::ConflictingChildLabel`]): rotating the one
+    /// picked would leave the other name live, so the walk aborts instead.
+    fn record_children(&self, body: &ReadBody) -> Result<Vec<[u8; 16]>, ResolveFailure> {
+        let ReadBody::Folder { children, .. } = body else {
+            return Ok(Vec::new());
+        };
+        let mut inner = self.inner.borrow_mut();
+        let mut ids = Vec::with_capacity(children.len());
+        for child in children {
+            if inner.child_scopes.contains(&child.id) {
+                continue;
+            }
+            let name = scope_name(&child.ipns_name)?;
+            match inner.names.entry(child.id) {
+                Entry::Occupied(seen) if *seen.get() != name => {
+                    return Err(ResolveFailure::ConflictingChildLabel);
+                }
+                Entry::Occupied(_) => {}
+                Entry::Vacant(slot) => {
+                    slot.insert(name);
+                }
+            }
+            ids.push(child.id);
+        }
+        Ok(ids)
+    }
 }
 
 /// The scope root's gated read, held across a failed publish.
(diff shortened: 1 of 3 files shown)


FSM1:
Applied the nitpick, at a different site than suggested.

The asymmetry is real and worth recording: \`ResolveFailure::ConflictingChildLabel\` is documented at its home (\`rotation/eager_set.rs\`) as *"Retryable: converges once the write-rotation re-point wave repairs both parent indexes"*, and \`CascadeError::is_retryable\`/\`SweepError::is_retryable\` both honour that. The write wave maps it to \`Rejected\` instead — which is correct, not a defect: this **is** the re-point wave, so it has nothing left to wait for, and retrying would spin on a conflict only this run can clear.

Documented at `wave_read_verdict` in `crates/engine/src/net/rotation.rs` rather than on `WriteSubtreeResolver::resolve_node`, because that function is where the classification decision actually lives. The trait method returns a `ResolveFailure` and classifies nothing; a reader chasing retry semantics from there lands on the error type, whose doc is what the write-wave case contradicts. Stating it at the decision site keeps the rationale in one place per the repo comment rules, instead of restating it on a caller.

`42018af0b`.

PR #1119, opened 2026-08-06
feat: convert invite claims into personal grants

### LINKED ISSUE #1015: engine: implement grant invites
## Already landed — do not re-implement

`crates/engine/src/grants/` is ~3,770 lines and the whole non-invite grants surface is written and simulation-tested: `create.rs`, `accept.rs`, `ledger.rs`, `child_index.rs`, `revocation.rs`, `owner_entry.rs`, `contact.rs`. **Grant creation is done; this issue does not touch it.**

**The first half of invites is also done.** The original scope claimed invites were "100% greenfield" with "no invite type, no ephemeral-key blob path, no claim handling, and no ledger expiry field". Three of those four are false as of `main` @ `3d087d673`:

- **The ephemeral-key grant blob path exists.** `crates/engine/src/grants/invite.rs` is 645 lines. `EphemeralInvitee` (`invite.rs:49`) derives the throwaway identity from one random invite secret exactly the way `SessionIdentity` derives a real one from a login secret — the secp256k1 scalar adopted directly, the X25519 sealing half through the frozen `enc-subkey` edge. `mint_invite_grant` (`invite.rs:154`) mints the invite's `GrantRow` through the same `mint_grant_row` every contact grant uses, and **derives** the scope root's `ipnsName` from `write_scope_seed` rather than accepting it as input. Both are exported from `crates/engine/src/grants/mod.rs:35`.
- **Byte-shape parity is established and tested.** On the envelope surface — blinded tag, commitment entry, and the blob `reseal_scope_root` wraps — an invite is shaped like a personal grantee's; the module's tests cover self-location, commitment enforcement, and the cross-scope/stranger negatives.
- **The ledger expiry field is landed.** `GrantLedgerEntry::expires_at: Option<NonZeroU64>` at `crates/core/src/seal/write_body.rs:64`, encoded/decoded with absence-preserving coverage in that module, and the liveness predicate `entry_is_live` at `crates/engine/src/grants/ledger.rs:79`.

So the original plan's PR **(a)** — "the ephemeral-key blob plus the ledger expiry field" — has shipped. Do not re-implement any of it.

## Scope

Only the original plan's PR **(b)** remains: **claim-to-grant conversion and bearer write-link flagging**, plus wiring the expiry predicate that today has no production caller.

1. **Claim-to-grant conversion.** The owner receives the sealed, ephemeral-key-signed claim over the mailbox, verifies the signature against the ephemeral public half committed in the ledger, and converts it into a personal grant for the claimant's contact-anchored identity. Multi-claim: one link yields N personal grants and the link itself stays live until it expires or is revoked. Nothing of this exists today — `crates/engine/src/grants/mod.rs` states that claim conversion is not implemented here, and `crates/engine/src/mailbox/mod.rs:46` notes only that the transport carries an opaque payload whose invite-claim meaning **the grants layer** must frame. `crates/core/src/payload/mailbox.rs` supplies the generic sealed envelope (`seal_mailbox_payload` / `open_mailbox_payload`) and nothing invite-specific.

   Note the property `invite.rs` documents on `EphemeralInvitee`: the invitee is structurally a login identity, so a ledger's `recipientIdentityPk` for an invite row is **not** a contact-anchored identity. Conversion is what re-anchors it, and the converted personal grant must be anchored to the claimant's imported contact, never to the ephemeral half.

2. **Expiry enforcement on the read path.** `entry_is_live` has no production caller at `3d087d673` — only `invite.rs`'s own test. An expired ledger entry must be inert wherever the read/accept path consults the ledger, before and independently of the discovered-link-expiry trigger pruning it. `GrantLedgerEntry::expires_at`'s own doc states what a deadline does and does not guarantee; honour that framing rather than treating expiry as a capability boundary.

3. **Bearer write-link flagging.** A write invite hands out extractable subtree signing keys, so the engine must flag write links to the host as bearer capabilities. Revoking or expiring a write link is only real via `rotate_scope_write`, never by pruning the ledger alone; the engine must say so at the API surface rather than letting a host believe a ledger prune revoked anything. `mint_invite_grant` already takes a `Permission`, so the minting side accepts a write invite today with no such flagging.

Only the owner may mint an invite, convert a claim, or revoke a link — the same owner-only authority rule as every commitment change. `mint_invite_grant` is owner-only by construction (it takes the owner's encryption subkey secret); conversion and revocation must enforce it explicitly.

**One PR.** The original two-PR split is spent: (a) is merged, and what is left is a single coherent slice.

## Blueprint anchors

`blueprint/engine.md`: Grants and ledger — Invites, Authority, Accept flow; Rotation primitives — Triggers (discovered link expiry). Read it; do not re-derive it.

`CONTEXT.md`: grant blob, grant ledger, grant-set commitment, blinded tag, writer pseudonym, structure signature, mailbox, contact code.

## Gate

Engine simulation. The byte-shape and self-location properties are already covered in `invite.rs` — do not duplicate them. New coverage:

- multi-claim: two independent claimants each convert to their own personal grant; the link stays live
- a converted grant is anchored to the claimant's contact-anchored identity, not to the ephemeral half
- a claim whose ephemeral signature does not verify against the committed public half is rejected fail-closed
- a claim against an expired ledger entry is refused, and an expired entry is inert on the read path before any prune runs
- a non-owner cannot convert or revoke
- write links are reported to the host as bearer capabilities, and revoking one without a write rotation is refused or reported as not-yet-revoked

## Depends on

None — grant creation, the ledger, the commitment, the invite blob path and the mailbox pipeline are all landed. This is deliberately **not** blocked on the rotation resolver/publisher wiring.

## Blocks

- the facade command arms — `CreateInviteLink` can mint a row today, but has no claim path to complete the flow

---

_Corrected against `main` @ `3d087d673` on 2026-08-06._ The prior body declared invites "100% greenfield". They are not: `crates/engine/src/grants/invite.rs` (645 lines, `EphemeralInvitee` at `:49`, `mint_invite_grant` at `:154`, exported at `grants/mod.rs:35`) and the ledger expiry field (`crates/core/src/seal/write_body.rs:64`, `entry_is_live` at `grants/ledger.rs:79`) had already landed, i.e. the whole of the original PR (a). Scope is now cut to the remaining work only, with expiry-predicate wiring added because `entry_is_live` has no production caller.

Part of #635

FSM1:
## What

Completes the invite slice. The ephemeral-key grant blob (`EphemeralInvitee`, `mint_invite_grant`) and the ledger expiry field already landed, so this PR is the remainder: claim-to-grant conversion, expiry enforcement, and bearer write-link flagging.

### Claim-to-grant conversion

- `InviteClaim` is the mailbox payload the grants layer frames — `{scopeRootName, contactCode}`, det-CBOR, opaque application bytes inside the seal.
- `post_invite_claim` seals it to the owner's encryption subkey and signs it, **as the mailbox sender**, with the link's ephemeral identity key. That is the blueprint's "sealed, ephemeral-key-signed mailbox request" verbatim: `open_mailbox_payload` verifies the inner sender signature fail-closed, and `poll_verified` drops anything that does not.
- `convert_invite_claim` binds that verified sender to a `RecordedInvite` — **owner-local state minted alongside the row** — imports the claimant's contact code fail-closed, and mints a personal grant through the same `mint_grant_row` every contact grant uses.
- The converted grant anchors to the **claimant's contact-anchored identity**, never the ephemeral half. A claim handing back the link's own self-signed contact code, or the owner's own bundle from the invite URL, is refused.
- Multi-claim: the link's commitment entry and ledger row survive conversion, so one link yields a grant per claimant until it expires or is revoked.
- Permission comes from the **owner-signed commitment entry**, never the write-grantee-mutable ledger row. A claimant who already holds a read grant and claims a write link is **upgraded** on both surfaces; a claim never lowers a committed permission, and `ConvertedClaim.row` can never contradict `ConvertedClaim.commitment`. `ClaimOutcome` reports which of the three happened.

### Why claimability comes from owner-local state

Both review gates independently found the same HIGH against the first draft, which located the link by scanning the resolved ledger for a row whose `recipientIdentityPk` matched the mailbox sender.

Nothing in a published record marks a row as an invite — that byte-shape parity is the design — and the fields that could say so sit **outside** the owner's signature (`enforce_committed_ledger` compares `(tag, permission)` only). So any committed grantee could post an honestly-signed claim naming a sockpuppet contact and have its own row convert, and a write-grantee could reach the same sink by re-authoring `recipientIdentityPk` on someone else's row. Either way the owner signs a grant for an identity it never approved, and that grant survives the rotation that revokes the attacker's own tag — a revocation-completeness break, not merely a re-share.

`RecordedInvite` fixes it at the root: conversion and revocation decide **what may be claimed** from the owner's own record and never from the record they act on. `revoke_invite_link` likewise cuts only a recorded link, so a mis-plumbed tag cannot silently revoke a personal grantee.

### Expiry

The owner's recorded deadline is the authority. The published `expiresAt` is honoured as well — an expired row is inert here, before any prune reaches it — but only ever to shorten a link, because a write-grantee can re-author that field. Both directions are pinned at the exact deadline instant, including a stripped published deadline failing to extend a link.

### Bearer write-link flagging

`mint_invite_grant` returns `MintedInvite { row, link, capability }`. `LinkCapability::BearerWrite` says the link hands out an extractable subtree signing key. `revoke_invite_link` reports `is_complete() == false` for a write link: the fragment holder keeps that key until `rotate_scope_write` runs, so a host can never read a ledger prune as a revocation.

### Fail-closed encode/decode symmetry

`check_publishable` is the produce-side mirror of what a resolver hard-rejects — the `MAX_GRANT_BLOBS` ceiling and duplicate tags (core rejects both at decode and inside `sign_grant_set`) and a ledger diverging from the commitment (the adoption gate's owner-authority check). It returns `Err`, release-active, and every arm is covered by a test that fires in a release build. The ceiling matters here specifically: a bearer link is multi-claim and contact codes are free to mint, so without it a link could walk a scope's grant set to the point where nothing can be signed.

## Review gates

`/simplify`, `/security-review` and `/crypto-privacy-review` all ran on `git diff main...HEAD`. Folded in: the HIGH above; the `MAX_GRANT_BLOBS` symmetry gap; the silent permission-upgrade drop; the owner-contact-as-claimant case; ambiguous-lookup refusal instead of first-match; `CommittedScope` must be the adopted record's set; `OwnerAuthority::identity_signer` documented as a capability token rather than a signing key; the claim replay contract; the claimant-to-owner linkage the transport learns; `InviteClaim`'s app-framing decode policy; distinct HPKE ephemeral scalars in the fixtures. `live_entry` was dropped — it had no consumer.

Filed rather than fixed here, both with a dependency edge on this issue:

- \#1127 — `reseal_scope_root` seals to `recipientEncPk` straight from the ledger with no tag re-derivation, so a write-grantee swapping that field redirects every future honest re-seal. Out of this PR's file ownership and strictly more severe than anything in the diff.
- \#1128 — an invite claim has no freshness, so a server-retained item redelivered after the owner cut the converted grant reads as a fresh claim. Needs a durable owner-side store this slice does not own; the caller contract is documented on `convert_invite_claim` meanwhile.

## Deliberately out of scope

- The rotation this drives. `revoke_invite_link` returns the cut plus the requirement; wiring `rotate_scope`/`rotate_scope_write` behind it belongs to the rotation slice, and `crates/engine/src/rotation/` and `grants/create.rs` are untouched.
- The discovered-link-expiry trigger, which consumes `entry_is_live` in its own slice.
- The facade `CreateInviteLink` / claim command arms.
- The `expiresAt`-outside-the-commitment residual raised on the part-(a) crypto review stands, still pinned by `a_write_grantee_may_strip_a_deadline_without_failing_owner_authority`. This PR adds no unattended prune, so the induced-revocation direction stays shut: nothing here re-signs a commitment because a deadline passed, and the owner's own recorded deadline is what gates a claim.

## Gate

`Engine Tests` (`cargo test -p cipherbox-engine`) — the unit suite in `grants::invite` plus a new `crates/engine/tests/invite_claims.rs` running the claim over the real mailbox seam. New coverage:

- two independent claimants each convert to their own personal grant; the link stays live
- a converted grant is anchored to the claimant's contact, not the ephemeral half
- only a link the owner recorded can be claimed — an ordinary grantee's row cannot, nor can a ledger row a write-grantee retargeted at an ephemeral identity
- a claim signed by a key the owner never recorded is refused fail-closed, as is one against a link cut from the committed set
- expiry at the exact deadline instant, from the owner's record and from the published row, with nothing pruned
- a non-owner can neither convert nor revoke; nor can the owner identity paired with a foreign encryption subkey
- a write link is flagged bearer and its ledger cut alone reports as not-yet-revoked
- upgrade, no-op and never-lower claim outcomes
- malformed claims and broken contact bindings
- the produce-side ceiling, duplicate-tag and divergence rejects
- the transport carries no claim field in the clear

## Gates run

| Command | Exit |
| --- | --- |
| `cargo fmt --all --check` | 0 |
| `node scripts/check-tracker-refs.mjs` | 0 |

Closes #1015

> [!NOTE]
> ### Convert invite claims into personal grants in the grants module
> - Adds `convert_invite_claim` to convert mailbox-delivered invite claims into personal `GrantRow` entries, enforcing expiry, owner authority, claimant identity, and upgrade semantics (read→write).
> - Adds `post_invite_claim` so invitees can post sealed, ephemeral-signed claims to the owner's mailbox via the `Mailbox` seam.
> - Adds `revoke_invite_link` to remove a link from the committed set, returning `InviteRevocation` which signals whether additional write-key rotation is needed.
> - Changes `mint_invite_grant` return type from `GrantRow` to `MintedInvite`, which bundles the network row with owner-local `RecordedInvite` metadata and a `LinkCapability` classification.
> - Adds `check_publishable` to enforce capacity (`MAX_GRANT_BLOBS`), tag uniqueness, and ledger authority before any set is published.
> - Risk: `mint_invite_grant` callers must update to access the grant row via `MintedInvite.row`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 576dedf.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added secure bearer-link invites with capability classification and owner-controlled lifecycle management.
  * Invitees can claim links and convert them into personal, contact-anchored grants.
  * Added invite expiry, revocation, authorization, duplicate detection, and validation.
  * Added secure claim delivery through the mailbox flow.

* **Bug Fixes**
  * Invalid, unauthorized, expired, malformed, or oversized claims now fail safely.

* **Tests**
  * Added end-to-end coverage for successful claims, authorization failures, expiry, revocation, and protected transport contents.

### DIFF
--- a/crates/engine/src/grants/invite.rs
+++ b/crates/engine/src/grants/invite.rs
@@ -25,8 +25,12 @@
 
 use cipherbox_core::codec::{Map, Value, decode, encode_fixed_depth};
 use cipherbox_core::error::{CodecError, Malformed};
-use cipherbox_core::seal::{GrantLedgerEntry, GrantSetCommitment, Permission, verify_grant_set};
-use cipherbox_core::suite::ecdsa::{EcdsaSignature, EcdsaSigner, EcdsaVerifier};
+use cipherbox_core::seal::{
+    GrantLedgerEntry, GrantSetCommitment, MAX_GRANT_BLOBS, Permission, verify_grant_set,
+};
+use cipherbox_core::suite::ecdsa::{
+    EcdsaSignature, EcdsaSigner, EcdsaVerifier, IDENTITY_PUBLIC_LEN,
+};
 use cipherbox_core::suite::secret::{SECRET_LEN, SecretBytes};
 use cipherbox_core::suite::x25519::{X25519Public, X25519Secret};
 use cipherbox_core::{ipns::IpnsName, kdf};
@@ -100,6 +104,13 @@ pub enum InviteError {
     /// Conversion exists to re-anchor a bearer link to a contact-anchored
     /// identity, so anchoring back to the ephemeral half is refused.
     ClaimantIsTheEphemeralHalf,
+    /// The claim handed back the owner's own contact bundle, which the invite URL
+    /// carries. The owner is not a grantee of its own scope.
+    ClaimantIsTheOwner,
+    /// The set is at the grant-set ceiling
+    /// ([`MAX_GRANT_BLOBS`](cipherbox_core::seal::MAX_GRANT_BLOBS)); one more row
+    /// could only ever mint a record its own decoder refuses.
+    GrantSetFull,
     /// The claimant's encryption subkey is non-contributory, so no blinded tag
     /// binds a grant to it.
     UnusableClaimantKey,
@@ -126,6 +137,8 @@ impl InviteError {
             Self::LinkExpired => "link-expired",
             Self::ClaimantContact(_) => "claimant-contact-invalid",
             Self::ClaimantIsTheEphemeralHalf => "claimant-is-the-ephemeral-half",
+            Self::ClaimantIsTheOwner => "claimant-is-the-owner",
+            Self::GrantSetFull => "grant-set-full",
             Self::UnusableClaimantKey => "unusable-claimant-key",
             Self::DuplicateTag => "duplicate-tag",
             Self::Authority(v) => v.check(),
@@ -226,12 +239,38 @@ impl LinkCapability {
     }
 }
 
-/// A minted invite link: the rows the owner commits, plus the capability flag
-/// the host renders and revocation must respect.
+/// One invite link as the owner recorded it at mint — **owner-local state, never
+/// network bytes**.
+///
+/// A published ledger row is deliberately byte-shaped like a personal grantee's,
+/// so nothing in a resolved record says "this row is an invite", and the fields
+/// that would say so (`recipientIdentityPk`, `expiresAt`) sit outside the owner's
+/// signature and are re-authorable by any write-grantee. Conversion therefore
+/// decides *what may be claimed* from this record and never from the record it
+/// converts against.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct RecordedInvite {
+    /// The link's blinded tag.
+    pub tag: [u8; 32],
+    /// The ephemeral identity a fragment holder signs its claim with.
+    pub ephemeral_identity_pk: [u8; IDENTITY_PUBLIC_LEN],
+    /// The ephemeral encryption subkey the link's blob is sealed to.
+    pub ephemeral_enc_pk: [u8; SECRET_LEN],
+    /// The deadline as minted. This copy is the authority: the published
+    /// `expiresAt` is a cooperating-reader hint a write-grantee can strip or
+    /// forge ([`GrantLedgerEntry::expires_at`]).
+    pub expires_at: Option<UnixMillis>,
+}
+
+/// A minted invite link: the rows the owner commits, the record the owner keeps,
+/// and the capability flag the host renders and revocation must respect.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct MintedInvite {
     /// The link's blinded tag, commitment entry and ledger row.
     pub row: GrantRow,
+    /// The owner-local record [`convert_invite_claim`] and
+    /// [`revoke_invite_link`] act on. Persist it with the link.
+    pub link: RecordedInvite,
     /// What the link hands out.
     pub capability: LinkCapability,
 }
@@ -252,7 +291,7 @@ pub fn mint_invite_grant(
     permission: Permission,
     expires_at: Option<UnixMillis>,
 ) -> Result<MintedInvite, InviteError> {
-    let expires_at = match expires_at {
+    let deadline = match expires_at {
         Some(deadline) => Some(NonZeroU64::new(deadline.0).ok_or(InviteError::InvalidExpiry)?),
         None => None,
     };
@@ -266,8 +305,14 @@ pub fn mint_invite_grant(
         permission,
     )
     .ok_or(InviteError::UnusableInviteeKey)?;
-    row.ledger_entry.expires_at = expires_at;
+    row.ledger_entry.expires_at = deadline;
     Ok(MintedInvite {
+        link: RecordedInvite {
+            tag: row.tag,
+            ephemeral_identity_pk: invitee.identity_pk().to_sec1(),
+            ephemeral_enc_pk: invitee.enc_public().to_bytes(),
+            expires_at,
+        },
         row,
         capability: LinkCapability::of(permission),
     })
@@ -298,7 +343,8 @@ impl InviteClaim {
     }
 
     /// Decode a claim (strict det-CBOR). A missing or mistyped field is
-    /// [`Malformed`].
+    /// [`Malformed`]. Unknown fields are dropped rather than preserved: this is a
+    /// consume-once engine payload, not a re-sealed shared structure.
     pub fn decode(bytes: &[u8]) -> Result<Self, CodecError> {
         let value = decode(bytes)?;
         let map = value.as_map()?;
@@ -318,11 +364,15 @@ impl InviteClaim {
 
 /// Post a claim to the owner's mailbox, sealed to the owner's encryption subkey
 /// and signed — as the mailbox sender — with the link's ephemeral identity key.
-/// That signature is what [`convert_invite_claim`] binds to the ephemeral public
-/// half the ledger commits, so only a fragment holder can claim.
+/// That signature is what [`convert_invite_claim`] binds to the ephemeral half the
+/// owner recorded, so only a fragment holder can claim.
 ///
 /// `owner` is the contact bundle the invite URL carries; `ephemeral_scalar` is
 /// fresh-per-call HPKE entropy from the injected seam.
+///
+/// Residual: the post is an authenticated API call addressed to the owner's
+/// identity key, so the transport learns claimant→owner even for claims the owner
+/// never converts. Inherent to the mailbox; the payload itself stays sealed.
 #[allow(clippy::too_many_arguments)]
 pub async fn post_invite_claim<M: Mailbox>(
     mailbox: &M,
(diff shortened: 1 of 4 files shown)


FSM1:
All three nitpicks applied in `1327e2c90`.

**`revoke_invite_link` tag re-derivation** — taken, and I'd rate it above defense-in-depth. The doc comment already claimed the rule ("a tag that is merely committed belongs to some grantee and is not this function's to cut") while the code only checked commitment membership. A doc asserting an invariant the code does not enforce is how the next change reasons wrongly, so the check now matches the comment, on the same rule `convert_invite_claim` applies at `:529`.

I did **not** take it on faith that the check works. The existing `unknown.link` assertion reaches `link-not-committed` through the commitment lookup, so it exercises nothing new. Added the case that isolates the re-derivation — a record carrying another grantee's *committed* tag with its own ephemeral key — then mutation-checked it: deleting the re-derivation block turns that assertion red (`26 passed; 1 failed`), restoring it returns `27 passed`.

**Ambiguous-link test** — added as suggested, `&[l.link, l.link]` asserting `link-not-committed`.

**Vacuity guard on the leak scan** — added. `windows()` yields nothing when the payload is shorter than the needle, so all three "no claim field in the clear" assertions could have passed without scanning anything. Guarded with `sealed.len() >= secret.len()` inside the loop rather than a single length assert, so each needle is covered individually.

Gates: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -D warnings`, the full `cipherbox-engine` suite, and `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown` all exit 0.

PR #1145, opened 2026-08-07
fix: fail closed on the transferred upload chunk and on an unknown worker command

### LINKED ISSUE #1037: web: zeroize the upload chunk in the engine worker after it is copied into WASM
`EngineHost.pushChunk` (`packages/client/src/worker/engineHost.ts`) receives the chunk by transfer, so the worker realm is its terminal owner. After `this.handle.pushChunk(handle, new Uint8Array(chunk))` copies the bytes into WASM memory, the JS-side plaintext survives until GC.

Surfaced by the crypto-privacy gate on the PR that made the follower upload path transfer-based end to end. The rest of that path now zeroes at every hop it owns — the follower transfers rather than clones, and the leader relay wipes a chunk it takes ownership of but cannot pass on. The worker is the one hop left.

## Scope

- Zero the transferred buffer immediately after the synchronous copy into WASM returns, and on the failure path too.
- Cover it with a test asserting a chunk handed to the host is zero afterwards.

AGENTS.md rule 7: clear sensitive material after use, at the terminal owner only.

### LINKED ISSUE #1051: web: fail closed in buildCommand instead of coercing wrong-typed command fields
`buildCommand` in `packages/client/src/worker/commandCodec.ts` switches over `CommandDescriptor` with no `default` arm, so an unknown `kind` falls out returning `undefined`, which reaches `wasm.command(undefined)` and throws in the wasm-bindgen glue. It fails closed, but by accident of the glue rather than by design — and the sibling switches at `commandCodec.ts:186` and `:214` already have fail-closed `default`s, so the pattern is inconsistent within the file.

The sharper case is a **known** kind carrying a wrong-typed field. `{ kind: 'rename', node: <16 bytes>, newName: 12345 }` reaches `wasm.Command.rename(nodeId, 12345)`, and wasm-bindgen's `String` marshalling coerces via `USVString` — renaming the node to `"12345"` rather than rejecting.

## Why it matters now

Surfaced by the crypto-privacy gate on #1039, which moved follower commands onto the private `MessagePort`. The relay there now rejects a payload that is not a command *shape* before forwarding, but deliberately leaves field validation to the codec: which commands exist, and what their fields must be, is the codec's to say. That division only holds if the codec actually enforces it.

Reachable from a version-skewed follower build with no hostile intent; a same-origin context can also drive it, though same origin is already the trust boundary, so the interesting case is skew rather than attack. No key or plaintext exposure — the mutation still goes through the engine's own authorization — but a user-visible metadata mutation with a silently coerced argument.

## Scope

- Give `buildCommand` a `default` arm binding the descriptor to `never` and throwing, matching the fail-closed `default`s already in the file.
- Reject wrong-typed fields rather than letting wasm-bindgen coerce them — at minimum the string and byte-array fields.
- Cover both with tests: an unknown `kind`, and a known `kind` whose string field is a number.

FSM1:
Two fail-closed fixes in the engine worker, in one PR. Both live in
`packages/client/src/worker/**`, touch disjoint files, and share one review lens,
so they are reviewed together rather than as two branches of a hundred lines each.

---

# Part 1 — #1037, wipe the transferred upload chunk

`EngineHost.pushChunk` receives the upload chunk by transfer, so the worker realm is its terminal owner — nothing upstream still holds the buffer. wasm-bindgen marshals the `Uint8Array` view into a `Vec<u8>` synchronously, before `push_chunk` returns its promise, so once the call has been made the JS-side copy is pure residue.

It was left for the collector. This wipes it.

## What changed

- `EngineHost` grows one private `scrubbing` helper: it runs a call over a view of the buffer and fills the view with zeroes in a `finally`, so the failure path scrubs too. `start` already did this inline for the login secret; `pushChunk` now shares the same helper rather than repeating the idiom.
- `EngineHostLike.pushChunk` states the ownership transfer, so an implementer knows the buffer is the host's to scrub.

The engine keeps its own copy in a `Zeroizing<Vec<u8>>` for the duration of the push, so scrubbing after the call settles rather than between the marshal and the await opens no new window.

## Tests

`packages/client/src/worker/engineHost.test.ts` gains two cases: a chunk handed to the host is all zeroes after a resolved push and after a rejected one, with the WASM-side copy asserted to have seen the real bytes so the wipe cannot pass by wiping too early.

## Not in this PR

The crypto-privacy pass found the same residue class on the sending side: `EngineClient.pushChunk` and `CorrelatedTransport.request` drop a chunk they refuse *before* any transfer, so it is never detached and never wiped. Those files are outside this batch's ownership — filed as #1155.

## Gates

`pnpm typecheck` 0, `pnpm test` 0, `pnpm lint` 0, `pnpm lint:tracker-refs` 0.

---

# Part 2 — #1051, fail closed on an unknown or wrong-typed command

`buildCommand` switched over `CommandDescriptor` with no `default` arm, so an unknown `kind` fell out returning `undefined` and only failed because the wasm-bindgen glue happened to reject it — fail-closed by accident, and inconsistent with the three sibling switches in the same file that already fail closed deliberately.

The sharper case was a **known** kind carrying a wrong-typed field. `{ kind: 'rename', node: <16 bytes>, newName: 12345 }` reached `wasm.Command.rename(nodeId, 12345)` and wasm-bindgen's `USVString` marshalling coerced it, renaming the node to `"12345"` instead of rejecting the command. Reachable from a version-skewed follower build: the private-port relay validates the command *shape* and deliberately leaves field validation to the codec, so that division only holds if the codec enforces it.

## What changed

- Every field the builders read is checked against the type the protocol declares — byte arrays must be a real `Uint8Array`, strings a real string, `opId` a real `bigint`. The checkers take `unknown`, because the descriptor arrives as plain data across a realm boundary and its static type is a claim, not a guarantee.
- `nodeKind` and `permission` no longer fall through to `Folder`/`Write` on an unrecognised literal; they reject it. That was the same silent-coercion class, one layer down.
- A `default` arm binds the descriptor to `never` and throws. Verified by temporarily adding a seventeenth kind to `CommandDescriptor`: `tsc` fails with `Type '{ kind: "probeNewKind"; … }' is not assignable to type 'never'` at that binding, so a new kind cannot be added without a builder.

Valid commands are unaffected — every check passes through the value it validated.

## Tests

`packages/client/src/worker/commandCodec.test.ts` gains five refusal cases driven against a permissive fake whose every builder succeeds, so only the codec's own checks can reject: an unknown kind, a numeric `newName` and a null `name`, a string where a public key belongs and an array where a node id belongs, an unrecognised `nodeKind` and `permission`, and a number `opId`. A sixth asserts the accept side of both mirror enums — `'folder'` and `'write'`, not only the literal each mapper tests first.

## Not in this PR

The byte and string arguments of `beginWrite`, `snapshot`, `download` and `openContentStream` reach WASM from the same untrusted message with the same coercion, and `passArray8ToWasm0` turns a 16-character string into sixteen zero bytes rather than rejecting it. That is #1154, which reuses the checkers this PR introduces.

## Gates

`pnpm typecheck` 0, `pnpm test` 0, `pnpm lint` 0, `pnpm lint:tracker-refs` 0.

---

Closes #1037
Closes #1051

> [!NOTE]
> ### Fail closed on invalid worker commands and transferred upload chunks
> - `buildCommand` in [commandCodec.ts](https://github.com/FSM1/cipher-box/pull/1145/files#diff-a2427e8ca2b1c7e229ffba009773c8fe15f5821745326cf1413e23275371319a) now strictly validates all descriptor fields, throwing descriptive errors for wrong-typed strings, non-`Uint8Array` byte fields, non-bigint `opId`, unknown node kinds, unknown permissions, and unknown command kinds instead of silently falling through or relying on wasm-bindgen coercion.
> - `EngineHost.pushChunk` in [engineHost.ts](https://github.com/FSM1/cipher-box/pull/1145/files#diff-25210c9231a2d0b9f72a4dc05fb431e4f740fe3dc44fe309b12877a29a5ef9f7) now zeros the provided `ArrayBuffer` after the WASM call completes, using a new `scrubbing` helper that guarantees zeroing even when the call rejects.
> - Behavioral Change: callers passing wrong-typed fields to `buildCommand` or unknown command kinds will now receive thrown errors rather than silent misbehavior.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 45f85cc.</sup>
> 
>

## Summary by CodeRabbit

- **Security & Reliability**
  - Strengthened command validation to reject unknown commands, invalid fields, unsupported values, and malformed operation IDs.
  - Upload data is now securely cleared from memory after processing, including when operations fail.
  - Updated upload behavior documentation to clarify that submitted plaintext buffers are scrubbed.

### DIFF
--- a/packages/client/src/worker/commandCodec.test.ts
+++ b/packages/client/src/worker/commandCodec.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from 'vitest';
 
 import { fakeWasmEnums } from '../testkit.js';
 import { buildCommand, readEvent, readSnapshot } from './commandCodec.js';
+import type { CommandDescriptor } from './protocol.js';
 import type { EngineWasm, WasmEvent, WasmSnapshotView } from './engineWasm.js';
 
 /**
@@ -53,6 +54,58 @@ describe('buildCommand', () => {
     buildCommand(wasm, { kind: 'cancelUpload', opId: 2n ** 60n });
 
     expect(calls).toEqual([[2n ** 60n]]);
+  });
+
+  /** Every builder succeeds, so only the codec's own checks can reject. */
+  const permissiveWasm = {
+    ...fakeWasmEnums,
+    NodeId: { fromBytes: (bytes: Uint8Array) => ({ bytes }) },
+    Command: new Proxy({}, { get: () => () => ({}) }),
+  } as unknown as EngineWasm;
+
+  const refuses =
+    (descriptor: unknown): (() => unknown) =>
+    () =>
+      buildCommand(permissiveWasm, descriptor as CommandDescriptor);
+
+  it('fails closed on an unknown command kind', () => {
+    expect(refuses({ kind: 'telepathy' })).toThrow('unknown command kind: telepathy');
+  });
+
+  it('rejects a wrong-typed string field rather than letting wasm-bindgen coerce it', () => {
+    expect(refuses({ kind: 'rename', node: new Uint8Array(16), newName: 12345 })).toThrow(
+      'invalid command field newName: number'
+    );
+    expect(
+      refuses({ kind: 'create', parent: new Uint8Array(16), name: null, nodeKind: 'file' })
+    ).toThrow('invalid command field name: null');
+  });
+
+  it('rejects a wrong-typed byte-array field', () => {
+    expect(
+      refuses({
+        kind: 'grant',
+        node: new Uint8Array(16),
+        recipientIdentityPublicKey: 'deadbeef',
+        permission: 'read',
+      })
+    ).toThrow('invalid command field recipientIdentityPublicKey: string');
+    expect(refuses({ kind: 'delete', node: [1, 2, 3] })).toThrow('invalid command field node');
+  });
+
+  it('rejects an unknown node kind or permission rather than defaulting one', () => {
+    expect(
+      refuses({ kind: 'create', parent: new Uint8Array(16), name: 'a', nodeKind: 'symlink' })
+    ).toThrow('invalid command field nodeKind: string');
+    expect(
+      refuses({ kind: 'createInviteLink', node: new Uint8Array(16), permission: 'admin' })
+    ).toThrow('invalid command field permission: string');
+  });
+
+  it('rejects an op id that is not the engine bigint', () => {
+    expect(refuses({ kind: 'cancelUpload', opId: 7 })).toThrow(
+      'invalid command field opId: number'
+    );
   });
 });
 
--- a/packages/client/src/worker/commandCodec.test.ts
+++ b/packages/client/src/worker/commandCodec.test.ts
@@ -56,6 +56,34 @@ describe('buildCommand', () => {
     expect(calls).toEqual([[2n ** 60n]]);
   });
 
+  it('maps the second literal of each mirror enum, not just the first', () => {
+    const calls: unknown[][] = [];
+    const record = (...args: unknown[]): object => {
+      calls.push(args);
+      return {};
+    };
+    const wasm = {
+      ...fakeWasmEnums,
+      NodeId: { fromBytes: (bytes: Uint8Array) => ({ bytes }) },
+      Command: { create: record, createInviteLink: record },
+    } as unknown as EngineWasm;
+
+    buildCommand(wasm, {
+      kind: 'create',
+      parent: new Uint8Array(16),
+      name: 'docs',
+      nodeKind: 'folder',
+    });
+    buildCommand(wasm, {
+      kind: 'createInviteLink',
+      node: new Uint8Array(16),
+      permission: 'write',
+    });
+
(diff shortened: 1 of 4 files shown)


FSM1:
## Nitpick disposition

Posted late — this PR merged without one. Verified against `origin/main`.

The review's single nitpick had two parts.

- **`packages/client/src/worker/engineHost.ts:29` — replace the behaviour narration with the plaintext-lifetime rationale. Taken**, in `45f85cc44`, the only commit after the review. It now reads "the host is its terminal owner, so it scrubs the plaintext to bound the lifetime of a copy no caller can reach".
- **`packages/client/src/worker/engineHost.test.ts:45` — delete the fake's docstring. Declined.** That comment carries the one non-obvious property of the fake: it forwards the *exact view* it was handed rather than a copy. The transfer and scrub assertions rest on that, and the name `pushingHost` does not convey it. Cosmetic either way.

Nothing outstanding.

PR #1148, opened 2026-08-07
fix: derive the cold-seed anchor role from authenticated scope identity

### LINKED ISSUE #775: engine: wire the shared-scope cold-seed caller with role from authenticated scope identity
## Context

#774 (Closes #763) added `gate::floor::AnchorRole { Root, Shared }` so the cold-seed read-epoch regression check runs only at the root/vault anchor. `AnchorRole::Shared` is currently **forward-scaffolding** — the only production `cold_seed_checked` caller (`sync/boot.rs`) passes `Root`; `Shared` is exercised only by unit tests.

## Invariant to preserve when the shared-scope cold-seed caller is wired

Flagged by both the #774 security and crypto-privacy reviews (INFO-1): when a production shared-scope cold-seed path is added (shared-folder / grant-accept boot), the caller **MUST** select `AnchorRole` from the scope's authenticated position in the vault-index walk (trusted state) — **never** from `RepointObject` contents or any network-supplied field.

Otherwise an untrusted transport could select `AnchorRole::Shared` for the actual vault anchor to suppress the read-epoch regression check. (It cannot lower the monotonic floor, so this degrades rollback *detection*/alerting rather than accepting a rollback — but the check must not be transport-selectable regardless.)

## Optional hardening (INFO-2)

Add a `boot::cold_start` integration test asserting `ColdStartError::FloorRegression` on a rolled-back vault repoint end-to-end (today the read-epoch fail-closed path is covered at the unit layer in `gate/floor.rs`, not through `cold_start`).

Refs #774, #763.

Part of #655

FSM1:
## What

`gate::floor::cold_seed_checked` took the `AnchorRole` as a caller-supplied
argument. The read-epoch (revocation) regression check runs only under
`AnchorRole::Root`, so any caller that could pick `Shared` for the vault anchor
would suppress that check. #774's security and crypto-privacy reviews flagged
this (INFO-1) as the invariant a future shared-scope cold-seed caller must
preserve.

This removes the choice rather than documenting it. The role is now derived
inside `cold_seed_checked` from the re-point's own `scope_id` against the
session's root scope id:

- `session_root_scope_id` is session state derived from the login secret.
- `RepointObject::scope_id` is sealed into the pointer-payload AAD
  (`pointer_ctx(v, object.scope_id)`) and covered by the owner's ECDSA
  signature over the object bytes, so a transport can neither forge it nor
  transplant a shared scope's re-point onto the vault anchor.

Neither operand is network-selectable, and `AnchorRole` is no longer nameable
by any caller — the invariant is structural, not a convention.

Also lands INFO-2: an end-to-end `boot::cold_start` test that seeds the floors
through a first real boot and then trips `ColdStartError::FloorRegression` on a
rolled-back but still validly owner-signed re-point at the same index. The
existing coverage pre-seeded the floor directly, so it never exercised the
seed-then-regress sequence.

## Not in this change

No production `AnchorRole::Shared` caller is wired, because the grantee-plane
shared-scope resolve does not exist yet: `sync::pointer::scope_pointer_name`
derives the scope pointer name from the **owner's** `ownerPointerSeed`, which a
grantee does not hold, and no wire field carries that name to a grantee.
`grants::accept` persists a `pointerReadKey` per share but nothing reads it
back, `sync::pointer::open_repoint` has no non-test caller, and
`Command::AcceptShare` still returns `EngineError::Unimplemented`. Wiring the
caller therefore means designing grantee-side pointer-name delivery, which is
its own slice. What this change guarantees is that when that slice lands, it
cannot get the role wrong.

## Gates

| Gate | Exit |
| --- | --- |
| `cargo fmt --all -- --check` | 0 |

## Review gates

`/simplify`, `/security-review` and `/crypto-privacy-review` all ran on
`git diff main...HEAD`. Folded in:

- **simplify (real):** `AnchorRole` became dead public surface once the role
  stopped being an argument. Inlined the comparison into `cold_seed_checked`
  and deleted the enum, its `impl`, and the unit test that asserted the private
  helper rather than behaviour. The read-epoch-soundness rationale is now stated
  once, on `cold_seed_checked`.
- **simplify (real):** the same rationale was restated four times. Now once.
- **simplify (nit):** the new end-to-end test overlapped
  `floor_regression_is_fail_closed_and_emits_nothing`. Merged into one test that
  seeds the floors through a real first boot and then replays a validly signed
  older re-point.
- **security MED / crypto-privacy MED, same mechanism:** the derivation rests on
  `RepointObject::scope_id` equalling the AAD scope its block was opened under,
  and only the encoder enforced that. Verified the mechanism: `seal_pointer_payload`
  derives the AAD from `object.scope_id`, and the AEAD tag covers the AAD, so a
  block whose inner `scopeId` disagreed with the AAD could not open at all. A
  decode-side equality check in `crates/core` would therefore be unreachable and
  could carry no reject vector, which the crate's reject-coverage law requires of
  every check. Pinned behaviourally instead: `a_repoint_opens_only_under_the_scope_id_it_names`
  in `crates/engine/tests/sync.rs`.
- **security LOW:** root scope ids are the constant `[0u8;16]` for every account,
  so scope ids are not globally distinguishing at the root. That is the defect
  tracked separately as the second-account epoch-floor lockout; not introduced
  here and not in scope for this change.

Closes #775

### DIFF
--- a/crates/engine/src/gate/floor.rs
+++ b/crates/engine/src/gate/floor.rs
@@ -118,15 +118,35 @@ impl std::error::Error for ColdSeedError {}
 /// never a security hole, since no rollback is accepted and the write-epoch
 /// check stays unconditionally sound. `Shared` therefore never runs the
 /// read-epoch check; only `Root` does.
+///
+/// The role is never an argument: [`cold_seed_checked`] derives it from the
+/// scope's authenticated identity, so no caller — and therefore no transport —
+/// can claim `Shared` for the vault anchor to suppress the read-epoch check.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum AnchorRole {
     /// The vault/root anchor: read-epoch check runs (alongside write-epoch).
     Root,
     /// A shared scope: read-epoch check skipped; only write-epoch runs.
-    ///
-    /// A production `Shared` caller MUST select this role from authenticated
-    /// scope identity, never a network/`RepointObject` field.
     Shared,
+}
+
+impl AnchorRole {
+    /// Select the role from authenticated scope identity: the session's own
+    /// vault anchor is [`AnchorRole::Root`], every other scope is
+    /// [`AnchorRole::Shared`].
+    ///
+    /// Both operands are trusted. `session_root_scope_id` is session state
+    /// derived from the login secret, and a [`RepointObject`]'s `scope_id` is
+    /// sealed into the pointer payload AAD and covered by the owner signature —
+    /// so an untrusted transport can neither forge a scope id nor transplant a
+    /// shared scope's re-point onto the vault anchor.
+    fn for_scope(session_root_scope_id: &[u8; 16], scope_id: &[u8; 16]) -> Self {
+        if scope_id == session_root_scope_id {
+            AnchorRole::Root
+        } else {
+            AnchorRole::Shared
+        }
+    }
 }
 
 /// Suffix that distinguishes a scope's write-epoch floor key from its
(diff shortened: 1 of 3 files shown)


FSM1:
Consolidated into #1157, which already contains both commits of this branch — #1157 was stacked on it, so its tip is a strict superset. Closing this to spend one review slot on the pair instead of two. No work is lost and no branch is deleted.

PR #1150, opened 2026-08-07
fix: fail closed on an unknown or wrong-typed command in the worker codec

### LINKED ISSUE #1051: web: fail closed in buildCommand instead of coercing wrong-typed command fields

FSM1:
`buildCommand` switched over `CommandDescriptor` with no `default` arm, so an unknown `kind` fell out returning `undefined` and only failed because the wasm-bindgen glue happened to reject it — fail-closed by accident, and inconsistent with the three sibling switches in the same file that already fail closed deliberately.

The sharper case was a **known** kind carrying a wrong-typed field. `{ kind: 'rename', node: <16 bytes>, newName: 12345 }` reached `wasm.Command.rename(nodeId, 12345)` and wasm-bindgen's `USVString` marshalling coerced it, renaming the node to `"12345"` instead of rejecting the command. Reachable from a version-skewed follower build: the private-port relay validates the command *shape* and deliberately leaves field validation to the codec, so that division only holds if the codec enforces it.

## What changed

- Every field the builders read is checked against the type the protocol declares — byte arrays must be a real `Uint8Array`, strings a real string, `opId` a real `bigint`. The checkers take `unknown`, because the descriptor arrives as plain data across a realm boundary and its static type is a claim, not a guarantee.
- `nodeKind` and `permission` no longer fall through to `Folder`/`Write` on an unrecognised literal; they reject it. That was the same silent-coercion class, one layer down.
- A `default` arm binds the descriptor to `never` and throws. Verified by temporarily adding a seventeenth kind to `CommandDescriptor`: `tsc` fails with `Type '{ kind: "probeNewKind"; … }' is not assignable to type 'never'` at that binding, so a new kind cannot be added without a builder.

Valid commands are unaffected — every check passes through the value it validated.

## Tests

`packages/client/src/worker/commandCodec.test.ts` gains five refusal cases driven against a permissive fake whose every builder succeeds, so only the codec's own checks can reject: an unknown kind, a numeric `newName` and a null `name`, a string where a public key belongs and an array where a node id belongs, an unrecognised `nodeKind` and `permission`, and a number `opId`. A sixth asserts the accept side of both mirror enums — `'folder'` and `'write'`, not only the literal each mapper tests first.

## Not in this PR

The byte and string arguments of `beginWrite`, `snapshot`, `download` and `openContentStream` reach WASM from the same untrusted message with the same coercion, and `passArray8ToWasm0` turns a 16-character string into sixteen zero bytes rather than rejecting it. That is #1154, which reuses the checkers this PR introduces.

## Gates

`pnpm typecheck` 0, `pnpm test` 0, `pnpm lint` 0, `pnpm lint:tracker-refs` 0.

Closes #1051

### DIFF
--- a/packages/client/src/worker/commandCodec.test.ts
+++ b/packages/client/src/worker/commandCodec.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from 'vitest';
 
 import { fakeWasmEnums } from '../testkit.js';
 import { buildCommand, readEvent, readSnapshot } from './commandCodec.js';
+import type { CommandDescriptor } from './protocol.js';
 import type { EngineWasm, WasmEvent, WasmSnapshotView } from './engineWasm.js';
 
 /**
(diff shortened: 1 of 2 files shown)


FSM1:
Consolidated into #1145. Both changes are fail-closed fixes in `packages/client/src/worker/**` on disjoint files, so they are reviewed together instead of costing two review slots for a hundred lines each. This branch's three commits were cherry-picked onto #1145 unchanged, and the combined branch passes `pnpm typecheck`, `pnpm test`, `pnpm lint` and `pnpm lint:tracker-refs`. Nothing is lost; this branch is not deleted.

PR #1157, opened 2026-08-07
fix: harden the epoch floor gates on cold seed and on revival

### LINKED ISSUE #703: engine: revival should corroborate recovered sequence against durable floor
From the retro `/security-review` of #692 (merged). **Severity: LOW.** Owner: rotation/revival slice (#635).

`crates/engine/src/net/revival.rs:63-83` — `revive()` fetches from a single authenticated `recovery_fetch(name)`, signature-verifies, and re-mints at `max(durable_floor, verified.sequence)+1`. Signature verification does not detect **replay** of an older but validly-signed record.

Failure scenario: a malicious/compromised recovery endpoint returns the user's own stale record (old head, low sequence) while the reviving device's durable floor is also stale (lost after the >EOL lapse). The device re-mints old content at a fresh higher sequence and silently rolls **itself** back to a prior legitimate state. Bounded — peers reject the lower sequence via the floor law, and the attacker is limited to the user's own past states (no forgery) — hence LOW.

Fix: cross-check the recovered sequence against the device's retained durable floor before accepting it as the CAS basis (the `max()` already covers the case where the local floor survived; the gap is when the floor is lost), and/or corroborate the recovered record against a `/routing/v1` fanout when any endpoint still serves the name.

Part of #655

### LINKED ISSUE #775: engine: wire the shared-scope cold-seed caller with role from authenticated scope identity

FSM1:
Two related floor-gate fixes in one PR. They were developed as a stack and share
`crates/engine/src/gate/floor.rs`, so they are reviewed together rather than as
two branches whose second one only ever applies on top of the first.

Read part 1 first — part 2's floor corroboration relies on the same durable
floor that part 1's derivation protects.

---

# Part 1 — #775, derive the cold-seed anchor role from authenticated scope identity

## What

`gate::floor::cold_seed_checked` took the `AnchorRole` as a caller-supplied
argument. The read-epoch (revocation) regression check runs only under
`AnchorRole::Root`, so any caller that could pick `Shared` for the vault anchor
would suppress that check. #774's security and crypto-privacy reviews flagged
this (INFO-1) as the invariant a future shared-scope cold-seed caller must
preserve.

This removes the choice rather than documenting it. The role is now derived
inside `cold_seed_checked` from the re-point's own `scope_id` against the
session's root scope id:

- `session_root_scope_id` is session state derived from the login secret.
- `RepointObject::scope_id` is sealed into the pointer-payload AAD
  (`pointer_ctx(v, object.scope_id)`) and covered by the owner's ECDSA
  signature over the object bytes, so a transport can neither forge it nor
  transplant a shared scope's re-point onto the vault anchor.

Neither operand is network-selectable, and `AnchorRole` is no longer nameable
by any caller — the invariant is structural, not a convention.

Also lands INFO-2: an end-to-end `boot::cold_start` test that seeds the floors
through a first real boot and then trips `ColdStartError::FloorRegression` on a
rolled-back but still validly owner-signed re-point at the same index. The
existing coverage pre-seeded the floor directly, so it never exercised the
seed-then-regress sequence.

## Not in this change

No production `AnchorRole::Shared` caller is wired, because the grantee-plane
shared-scope resolve does not exist yet: `sync::pointer::scope_pointer_name`
derives the scope pointer name from the **owner's** `ownerPointerSeed`, which a
grantee does not hold, and no wire field carries that name to a grantee.
`grants::accept` persists a `pointerReadKey` per share but nothing reads it
back, `sync::pointer::open_repoint` has no non-test caller, and
`Command::AcceptShare` still returns `EngineError::Unimplemented`. Wiring the
caller therefore means designing grantee-side pointer-name delivery, which is
its own slice. What this change guarantees is that when that slice lands, it
cannot get the role wrong.

## Gates

| Gate | Exit |
| --- | --- |
| `cargo fmt --all -- --check` | 0 |

## Review gates

`/simplify`, `/security-review` and `/crypto-privacy-review` all ran on
`git diff main...HEAD`. Folded in:

- **simplify (real):** `AnchorRole` became dead public surface once the role
  stopped being an argument. Inlined the comparison into `cold_seed_checked`
  and deleted the enum, its `impl`, and the unit test that asserted the private
  helper rather than behaviour. The read-epoch-soundness rationale is now stated
  once, on `cold_seed_checked`.
- **simplify (real):** the same rationale was restated four times. Now once.
- **simplify (nit):** the new end-to-end test overlapped
  `floor_regression_is_fail_closed_and_emits_nothing`. Merged into one test that
  seeds the floors through a real first boot and then replays a validly signed
  older re-point.
- **security MED / crypto-privacy MED, same mechanism:** the derivation rests on
  `RepointObject::scope_id` equalling the AAD scope its block was opened under,
  and only the encoder enforced that. Verified the mechanism: `seal_pointer_payload`
  derives the AAD from `object.scope_id`, and the AEAD tag covers the AAD, so a
  block whose inner `scopeId` disagreed with the AAD could not open at all. A
  decode-side equality check in `crates/core` would therefore be unreachable and
  could carry no reject vector, which the crate's reject-coverage law requires of
  every check. Pinned behaviourally instead: `a_repoint_opens_only_under_the_scope_id_it_names`
  in `crates/engine/tests/sync.rs`.
- **security LOW:** root scope ids are the constant `[0u8;16]` for every account,
  so scope ids are not globally distinguishing at the root. That is the defect
  tracked separately as the second-account epoch-floor lockout; not introduced
  here and not in scope for this change.

---

# Part 2 — #703, corroborate the revival basis against the fan-out and the durable floor

## The defect

`net::revival::revive` fetched from a single authenticated
`ApiClient::recovery_fetch`, verified the signature, and used the recovered
sequence as the CAS basis. Signature verification proves authorship, not
freshness, so a malicious or compromised recovery endpoint could serve the
user's own older record. The device would then re-mint that superseded head CID
at a fresh, higher sequence and roll itself back to a prior legitimate state —
and, because the sequence is fresh, peers would accept the rollback under the
floor law rather than reject it.

## The fix

Two corroborations before the basis is accepted.

1. **Fan-out corroboration.** `fanout_get_verify` is consulted alongside the
   recovery endpoint; whichever authenticated record has the higher sequence
   becomes the basis. Crucially the head CID is now read from the *same record*
   as the sequence, so a corroborated sequence can never be paired with a stale
   CID — taking `max()` of the sequences alone would have fixed the CAS number
   while still republishing superseded content.
2. **Durable floor corroboration.** The device's durable sequence floor is what
   it last adopted for the name. A basis strictly below it is a demonstrably
   rolled-back source — a replay, not a lapse — and now fails closed with
   `ReviveError::StaleSource { floor, recovered }` before anything is registered
   or PUT.

A floor-read failure is `ReviveError::FloorRead` — host I/O, never a trust
verdict, but the corroboration cannot run without it, so it is not swallowed.

## Notes

- Behaviour is unchanged in the ordinary lapse case: when no endpoint still
  serves the name the fan-out returns nothing and the recovery record is the
  basis, exactly as before. Both existing revival tests and the multi-day EOL
  timeline pass untouched.
- Determinism is preserved: no clock or RNG is read on the new path.
- `revive` has no production caller yet; it is exported engine API, so the
  hardening lands with the slice rather than after it is wired.

## Tests

- `revive_prefers_a_fresher_corroborating_record_over_the_recovery_endpoint` —
  a replaying recovery endpoint at sequence 3 against a live endpoint at 9;
  asserts the re-mint lands at 10 **and** carries the fresher CID.
- `revive_fails_closed_when_every_source_is_below_the_durable_floor` — floor 9,
  recovery replays 3; asserts `StaleSource` and that no endpoint holds a record.

## Gates

| Gate | Exit |
| --- | --- |
| `cargo fmt --all -- --check` | 0 |

## Review gates

`/simplify`, `/security-review` and `/crypto-privacy-review` all ran on this
PR's own diff. Folded in:

- **security LOW:** a same-sequence fork between the two sources is a
  designed-for state — `publish` retries an `Unconfirmed` publish at the same
  sequence — and the strict `>` let the recovery endpoint pick which side of one
  was re-minted. The routing set is the canonical plane, so it now takes the tie
  at `>=`.
- **simplify (real) x4:** the module header still described the pre-change
  single-source flow; the corroboration and floor-law rationale were each stated
  three times; `StaleSource.recovered` named a value that may come from the
  fan-out; `FloorRead`'s doc contradicted `PublishError::FloorRead`. Header
  corrected, rationale stated once at its home, field renamed to `sequence` to
  match `RejectionReason::SequenceNotNewer`, doc reduced to a cross-reference.
- **simplify + crypto-privacy (test gaps):** the fan-out-is-older case and the
  fail-closed floor-read case were both untested. Added, plus two shared
  endpoint-assertion helpers that remove the duplicated loops.
- **crypto-privacy INFO:** verified the basis chain end to end — every candidate
  is verified against the Ed25519 key extracted from the name itself, with the
  sequence and value read from the same signed `data` blob, so a hostile endpoint
  can only replay authentic records.

Deliberately **not** folded in, filed as #1161 with a `blocked_by` edge on this
issue:

- `fanout_get_verify` cannot distinguish "no endpoint answered" from "no endpoint
  has this name", so the corroboration silently degrades to the recovery endpoint
  alone when the routing set is unreachable. Fixing it changes a return type
  consumed by `resolve`, `publish`, `settings`, `rotation` and `pointer_fetch`.
- A basis exactly at the durable floor is admitted with any content, so revival
  can still re-mint pre-rotation content when another device published a
  sequence this one never adopted. Closing it needs the adopted head value
  retained alongside the sequence floor, or the basis put through the adoption
  gate.

Neither is live: `revive` has no production caller yet.

---

Closes #775
Closes #703

> [!NOTE]
> ### Harden epoch floor gates on cold seed and revival against stale or rolled-back sources
> - Replaces the `AnchorRole` enum in `cold_seed_checked` with a `session_root_scope_id` parameter; the read-epoch check now fires automatically when `repoint.scope_id` matches the root scope, removing caller-controlled opt-out.
> - `revive` corroborates the recovery endpoint record against the routing fan-out via `fanout_get_verify`, preferring the fresher or equal-sequence fan-out record as the authoritative basis.
> - `revive` now reads the durable sequence floor and rejects any basis below it with `ReviveError::StaleSource`; floor read failures surface as `ReviveError::FloorRead` (fail-closed).
> - New integration tests in [`tests/net.rs`](https://github.com/FSM1/cipher-box/pull/1157/files#diff-f1714fdc0a1863958ce3776d7af837cca5f56a01304736850bb78aec9a02f23d) and [`tests/sync.rs`](https://github.com/FSM1/cipher-box/pull/1157/files#diff-139eb1e1e705dd3e78bb9e7a763382df09a2a1276b351a2be222cb3cc3842cd8) cover corroboration preference, fork resolution, stale-floor rejection, and rolled-back re-point rejection.
> - Behavioral Change: shared scopes no longer trigger the read-epoch regression check; callers can no longer bypass the check by passing a role selector.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c1d39ab.</sup>
> 
>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery to select the freshest valid routing or recovered record.
  * Prevented stale recovery sources from being accepted or published.
  * Added fail-closed handling when durable sequence information cannot be read.
  * Strengthened scope validation so sealed re-points only open under their authorized scope.
  * Prevented read- and write-epoch regressions during initialization and recovery.

### DIFF
--- a/crates/engine/src/gate/floor.rs
+++ b/crates/engine/src/gate/floor.rs
@@ -118,15 +118,35 @@ impl std::error::Error for ColdSeedError {}
 /// never a security hole, since no rollback is accepted and the write-epoch
 /// check stays unconditionally sound. `Shared` therefore never runs the
 /// read-epoch check; only `Root` does.
+///
+/// The role is never an argument: [`cold_seed_checked`] derives it from the
+/// scope's authenticated identity, so no caller — and therefore no transport —
+/// can claim `Shared` for the vault anchor to suppress the read-epoch check.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum AnchorRole {
     /// The vault/root anchor: read-epoch check runs (alongside write-epoch).
     Root,
     /// A shared scope: read-epoch check skipped; only write-epoch runs.
-    ///
-    /// A production `Shared` caller MUST select this role from authenticated
-    /// scope identity, never a network/`RepointObject` field.
     Shared,
+}
+
+impl AnchorRole {
+    /// Select the role from authenticated scope identity: the session's own
+    /// vault anchor is [`AnchorRole::Root`], every other scope is
+    /// [`AnchorRole::Shared`].
+    ///
+    /// Both operands are trusted. `session_root_scope_id` is session state
+    /// derived from the login secret, and a [`RepointObject`]'s `scope_id` is
+    /// sealed into the pointer payload AAD and covered by the owner signature —
+    /// so an untrusted transport can neither forge a scope id nor transplant a
+    /// shared scope's re-point onto the vault anchor.
+    fn for_scope(session_root_scope_id: &[u8; 16], scope_id: &[u8; 16]) -> Self {
+        if scope_id == session_root_scope_id {
+            AnchorRole::Root
+        } else {
+            AnchorRole::Shared
+        }
+    }
 }
 
 /// Suffix that distinguishes a scope's write-epoch floor key from its
@@ -309,7 +329,10 @@ pub async fn cold_seed<F: FloorStore>(floors: &F, repoint: &RepointObject) -> Se
 /// regresses does it advance the floors via the monotonic-max [`cold_seed`].
 ///
 /// The read-epoch check runs under [`AnchorRole::Root`] only; the write-epoch
-/// check is unconditional — see [`AnchorRole`] for why.
+/// check is unconditional — see [`AnchorRole`] for why. The role is derived from
+/// the re-point's own scope id against `session_root_scope_id` (the session's
+/// vault anchor, from the derived session identity), never supplied by the
+/// caller.
 ///
 /// Check-then-advance is deliberately not a CAS pair: the engine is the single
 /// writer (blueprint/engine.md "Facade"), and the monotonic-max store backstops
@@ -317,9 +340,9 @@ pub async fn cold_seed<F: FloorStore>(floors: &F, repoint: &RepointObject) -> Se
 pub async fn cold_seed_checked<F: FloorStore>(
     floors: &F,
     repoint: &RepointObject,
-    role: AnchorRole,
+    session_root_scope_id: &[u8; 16],
 ) -> Result<(), ColdSeedError> {
-    if role == AnchorRole::Root {
+    if AnchorRole::for_scope(session_root_scope_id, &repoint.scope_id) == AnchorRole::Root {
         if let Some(floor) = read_epoch_floor(floors, &repoint.scope_id)
             .await
             .map_err(ColdSeedError::Seam)?
@@ -368,6 +391,9 @@ mod tests {
     use crate::testkit::fakes::InMemoryFloorStore;
 
     const SCOPE: [u8; 16] = [7u8; 16];
+    /// A scope id that is not the session's vault anchor — cold-seeding it is a
+    /// shared-scope seed.
+    const SHARED_SCOPE: [u8; 16] = [8u8; 16];
     const NAME: &[u8] = b"k51-scope-root-name";
 
     #[test]
@@ -454,13 +480,13 @@ mod tests {
         let floors = InMemoryFloorStore::default();
         block_on(async {
             // A fresh (unseeded) scope adopts the owner-vouched anchor.
-            cold_seed_checked(&floors, &repoint(SCOPE, 5, 3), AnchorRole::Root)
+            cold_seed_checked(&floors, &repoint(SCOPE, 5, 3), &SCOPE)
                 .await
                 .unwrap();
             assert_eq!(read_epoch_floor(&floors, &SCOPE).await.unwrap(), Some(5));
             assert_eq!(write_epoch_floor(&floors, &SCOPE).await.unwrap(), Some(3));
             // Re-seeding at or above the floor advances monotonically.
-            cold_seed_checked(&floors, &repoint(SCOPE, 7, 3), AnchorRole::Root)
+            cold_seed_checked(&floors, &repoint(SCOPE, 7, 3), &SCOPE)
                 .await
                 .unwrap();
             assert_eq!(read_epoch_floor(&floors, &SCOPE).await.unwrap(), Some(7));
@@ -471,11 +497,11 @@ mod tests {
     fn cold_seed_checked_rejects_a_read_epoch_regression_fail_closed() {
         let floors = InMemoryFloorStore::default();
         block_on(async {
-            cold_seed_checked(&floors, &repoint(SCOPE, 5, 3), AnchorRole::Root)
+            cold_seed_checked(&floors, &repoint(SCOPE, 5, 3), &SCOPE)
                 .await
                 .unwrap();
             // A rolled-back re-point vouching a lower minReadEpoch is fail-closed.
-            let err = cold_seed_checked(&floors, &repoint(SCOPE, 4, 3), AnchorRole::Root)
+            let err = cold_seed_checked(&floors, &repoint(SCOPE, 4, 3), &SCOPE)
                 .await
                 .expect_err("read-epoch regression is a trust violation");
             assert_eq!(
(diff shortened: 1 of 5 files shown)


FSM1:
## Nitpick disposition

Posted late — this PR merged without one. Verified against `origin/main`. **Both nitpicks were taken pre-merge**, in `c1d39abab`.

- **`crates/engine/tests/net.rs:114, :134` — drop the docs restating the helper names.** Both `///` lines are gone; the helpers are bare on `main`.
- **`crates/engine/tests/net.rs:1101` — cover a same-sequence fork where the recovery and fan-out records carry different CIDs.** `revive_takes_the_fan_out_side_of_a_same_sequence_fork` exists and asserts `PublishOutcome::Published { sequence: 6 }`, all endpoints at 6, and the fan-out CID as the head value. That is the tie-break this PR hardened, so it is a real gate rather than a cosmetic addition.

Nothing outstanding.

PR #1160, opened 2026-08-07
feat: give the media player a surface for a recoverable stream refusal

### LINKED ISSUE #1064: web: the media player has no surface for a recoverable stream refusal

FSM1:
## What

`EngineError::TooManyStreams` is reachable only through the media pipe, and the pipe had nowhere to say so. `MediaBroker` handles the ceiling itself — it gives back the streams held only by the pin linger and retries the refused open once — but if that retry also fails, the body errors and the media element sees a bare network failure with no explanation and no affordance.

There was also no player: audio and video were not preview kinds at all, so the surface had no host.

## How the reason reaches the UI

A `ReadableStream` error on a response body reaches a media element as an untyped network failure — the code cannot ride the body out to the `<video>`. So it travels tab-side, where the broker already is:

- `MediaBroker` takes an `onFailure` seam and reports every read it gives up on as a `MediaFailure` carrying the engine's code (`engineErrorCode`), not a message string.
- `MediaService.onStreamError(listener)` classifies that with `isRecoverableEngineError` and hands the holder of the stream URL a `MediaStreamFailure` — `{ url, message, recoverable }`. It is the same classifier the vault browser already uses for a snapshot pull, so a ceiling and a verdict are told apart in exactly one place.

## The player

`MediaPlayer` renders the element over the ticket URL and subscribes to failures on that URL:

- **recoverable** — a notice plus a `retry` that remounts the element against the same ticket. The pipe refused a read; it did not withdraw the capability, so the ticket is still live and nothing is re-minted.
- **fault** — the engine's own message, terminal, no retry.
- The element's own `error` event is the floor for a failure the pipe never named (an unsupported container, a decode refusal). A classified refusal that already landed wins over it.

Audio and video join `previewKind`, streamed range by range with **no buffered fallback**: decoding a video whole into the tab is precisely what the pipe exists to avoid, and without a controlling Service Worker the preview refuses and points at download.

## Deviation from the issue

The issue expects `cb:media:error` to carry the engine's code across the port. Re-checked against the tree: that would be a dead field. The only `cb:media:error` that carries an engine code is raised by `MediaBroker.fail`, which runs **in the tab** — the same context as the player — and the round trip through the Service Worker can only lose information, because the pipe's sole action on it is to error a body whose error the media element cannot read. `MediaPipe`'s own failures ('media pull timed out', 'media port replaced') have no engine code to carry. So the code is reported tab-side and the wire is left alone.

## Gate

`packages/client/src/media/service.test.ts` drives two real refusals through the broker — `tooManyStreams` and `trustViolation` — and asserts each is reported against the URL it refused on with the right `recoverable` verdict, and that unsubscribing stops it. `FileBrowserActions.test.tsx` asserts the video plays off a ticket without a facade read, that a ceiling refusal renders a retry that keeps the ticket, that a fault does not, that another stream's refusal is ignored, and that the element's own error still reports.

## Verification

`pnpm typecheck` 0, `pnpm test` 0, `pnpm lint` 0, `pnpm lint:tracker-refs` 0. No Rust surface in this change.

Review gates: `/simplify` and `/security-review` both run on `git diff main...HEAD`. Security found nothing at or above the bar — the report carries no plaintext or key material, `safeMimeType` plus `nosniff` and the `no-store` clamp still bound what a ticket can render as, the classifier is a single-code equality so nothing widens into it, and the listener unsubscribes on unmount and on an engine rebuild. Its one nit and the `/simplify` findings are folded into the second commit: the broker classifies rather than round-tripping a raw code, the fan-out reuses `fanOut` so a throwing subscriber cannot drop the event, `onStreamError` refuses a subscription after `dispose`, `StreamRegistry` owns the ticket-URL shape, the three streamed preview branches became one, and the player reuses the vault browser's existing recoverable-refusal notice instead of a second dialect.

Runtime verification is not reachable from a unit harness — the surface needs a logged-in vault, a controlling Service Worker, and the engine at its open-stream ceiling. To check it by hand: bring up the local stack, upload an `.mp4`, open its preview, confirm playback streams over `/stream/<ticket>` with no facade `download`, then hold `MAX_OPEN_STREAMS` open from other tickets and seek — the notice with `[retry]` should appear over the player and the retry should recover once a stream frees.

Closes #1064

### DIFF
--- a/apps/web/src/components/file-browser/FileBrowserActions.test.tsx
+++ b/apps/web/src/components/file-browser/FileBrowserActions.test.tsx
@@ -3,6 +3,7 @@ import {
   type EngineClient,
   type EventDescriptor,
   type MediaService,
+  type MediaStreamFailure,
   type SnapshotDescriptor,
 } from '@cipherbox/client';
 import { act, fireEvent, render, screen, waitFor } from '@testing-library/react';
@@ -742,11 +743,18 @@ describe('the vault browser read path over the streaming pipe', () => {
   const revoked: string[] = [];
   const clicked: { href: string | null; download: string }[] = [];
   const originalClick = HTMLAnchorElement.prototype.click;
+  const streamListeners = new Set<(failure: MediaStreamFailure) => void>();
+
+  /** Stands in for the broker giving up on a read, which is always tab-side. */
+  const reportStreamError = (failure: MediaStreamFailure): void => {
+    for (const listener of streamListeners) listener(failure);
+  };
 
   beforeEach(() => {
     minted.length = 0;
     revoked.length = 0;
     clicked.length = 0;
+    streamListeners.clear();
     mediaControl.create = () =>
       ({
         streaming: true,
@@ -755,6 +763,10 @@ describe('the vault browser read path over the streaming pipe', () => {
         createStreamUrl: (source: { node: Uint8Array; size: number; mimeType: string }) => {
           minted.push(source);
           return `/stream/ticket-${minted.length}`;
+        },
+        onStreamError: (listener: (failure: MediaStreamFailure) => void) => {
+          streamListeners.add(listener);
+          return () => streamListeners.delete(listener);
         },
         revokeStreamUrl: (url: string) => {
           revoked.push(url);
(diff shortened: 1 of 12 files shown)


FSM1:
Consolidated into #1146. These two branches conflicted with each other in `apps/web/src/components/file-browser/FileBrowserActions.test.tsx`, where both add a method to the same hand-rolled `MediaService` fake, and they share `broker.ts`, `service.ts` and `service.test.ts`. Combining them resolves that conflict now — keeping both sides — instead of leaving it for whichever merged second. This branch's two commits were cherry-picked onto #1146; the resolved `service.ts` carries both `onStreamError` and `whenStreamIdle`, and the combined branch passes `pnpm typecheck`, `pnpm test` at 257 web tests, `pnpm lint` and `pnpm lint:tracker-refs`. Nothing is lost; this branch is not deleted.

PR #1190, opened 2026-08-08
fix: mint the write-plane history link and recover a resumed wave from published state

### LINKED ISSUE #1139: engine: a resumed write rotation cannot recover its write scope seed, because nothing mints the write-plane history link
## Problem

Nothing in the repo ever mints a `WriteBody.write_history_link`. `reseal_scope_root` copies whatever it is handed (`crates/engine/src/rotation/reseal.rs:467`), and every call site passes `&[]` or passes the old value through — including `WriteWaveNet::reseal_root` (`crates/engine/src/net/rotation.rs:1232`).

A resumed wave instead takes the fresh write scope seed from `RotateScopeWritePlan::resume_write_scope_seed` (`crates/engine/src/rotation/rotate_write.rs:246`), whose own doc says the recovery wiring is not landed and is faked in tests. The module header repeats this at `rotate_write.rs:38`.

## Why it matters

The crash-recovery tests are real — `mid_wave_crash_resumes_from_published_records_only` (`rotate_write.rs:1243`) and `a_resumed_wave_re_resolves_the_subtree_from_published_records_alone` (`net/rotation.rs:3864`) both pass. They pass because the test hands the seed back in memory.

A real crash has no published source for the seed. The seed is only reachable from the new root, and the new root's name is derived from the seed itself. So after a genuine mid-wave crash the wave cannot resume, and the subtree is stranded half-moved.

#635's own gate names "resumed name waves via the write-plane history link". This is an unmet slice gate, not a nicety.

## Proposed fix

Mint the write-plane history link at the root re-seal so a resumed wave recovers the fresh write scope seed from published state alone. Keep the seam boundary at entropy-versus-published-record, never an in-memory carry.

Related: #764 names the write-plane history link as the mechanism that makes orphaned interior names recoverable, so it depends on this.

### LINKED ISSUE #1179: engine: the write wave carries a writer-authored writeHistoryLink into the owner-signed moved root
## Problem

`WriteWaveNet::reseal_root` carries `plane.write_body.write_history_link` verbatim into the moved root's re-sealed write body (`crates/engine/src/net/rotation.rs`, the `CommittedSet` it hands `reseal_scope_root`).

`WriteBody` is sealed under `writeScopeSeed`, which **every** write grantee holds — including the one the rotation is cutting. `write_history_link` is opaque bytes (`crates/core/src/seal/write_body.rs`), no owner signature covers it, `enforce_committed_ledger` does not reach it, and `reseal_scope_root` copies it unexamined.

## Why it matters

Two compounding problems:

1. **Semantic wrong-seed.** The wave mints a *fresh* `writeScopeSeed` but republishes the *pre-rotation* write history link, so the moved root's link does not chain to the seed it was just re-keyed to.
2. **Capability widening once seed recovery lands.** `RotateScopeWritePlan::resume_write_scope_seed` is documented as "the fresh write scope seed recovered from the published root (the write-plane history link)" (`crates/engine/src/rotation/rotate_write.rs`), and that recovery wiring is not landed. When it lands, a revoked write grantee who plants a forged `writeHistoryLink` before the wave gets the **owner to sign it into the new root**, and the resumed wave then republishes the whole subtree under names derived from an attacker-chosen seed. Total write-plane takeover, in a record that is owner-authentic — no gate stage can catch it.

The only thing keeping this latent is that the recovery is unbuilt.

## Resolving

The wave must mint its own link rather than carry one: either pass an empty `write_history_link` from `reseal_root` (the wave holds its own seed; recovery is not this record's job), or extend `ResealSeeds` with a write-plane `prev` and have `reseal_scope_root` seal a fresh link over the retiring `writeScopeSeed` under `write_key(fresh_seed)`, mirroring the read-plane `prev` path.

Independently, when the seed recovery lands it must cross-check `derive_write_name(recovered_seed, scope_id) == published root name` before a single republish — the reverse of the check `publish_moved` already makes in the forward direction.

## Gate

Engine unit:

- a planted `writeHistoryLink` on the pre-wave root is never re-signed into the moved root
- a resumed wave refuses a recovered write scope seed that does not derive the published root name

## Notes

Found by the `/security-review` gate on #1177. Pre-existing; the fix lands in `crates/engine/src/rotation/reseal.rs` and `rotate_write.rs`, owned by other slices.

Part of #655

FSM1:
Two halves of one change. Landing the recovery without the guard would be a write-plane takeover, so they ship together.

`WriteWaveNet::reseal_root` carried the writer-authored `writeHistoryLink` verbatim into the owner-signed moved root. `WriteBody` is sealed under `writeScopeSeed`, which **every** write grantee holds — including the one the rotation is cutting — and no owner signature covers that field. Nothing minted the field either, so a genuine mid-wave crash had no published source for the fresh write scope seed and the subtree stranded half-moved; the existing crash-recovery tests passed only because the test handed the seed back in memory.

## Design choice

Of the two options #1179 sketches, this takes the second — extend `ResealSeeds` with a write-plane `prev` and mint the link — because it is the one that also satisfies #1139, and because the empty-link option would have deleted a field the blueprint makes load-bearing (`blueprint/api.md`: "interior old names are batch-retired at wave completion — a resumed wave enumerates them via the write-plane history link").

The shape is a two-variant enum rather than an extra field, so the invariant is structural rather than a runtime guard:

```rust
pub enum WriteHistory<'a> {
    Carried(&'a [u8]),      // the write plane is untouched
    Cut(PrevEpochSeed<'a>), // the write plane is being cut
}
```

`CommittedSet.write_history_link` is removed, so **a cut has no channel to carry through** — it is unrepresentable rather than checked. `Carried` keeps the existing byte-for-byte behaviour on every path that leaves the write plane alone (read rotation, sweep, cascade, grant create), so no live carry is dropped.

The link mirrors the read plane exactly: `seal_history_link` under `kdf::structure_key(fresh_write_scope_seed, history-link)`, AAD at the **new write epoch**, payload = the **retiring** seed and epoch. That is a use of the existing frozen `structure-key` edge — `blueprint/core.md`'s catalog row reads "nodeSeed **or scope seed**, structTag" — not a new edge, so no KAT regeneration and no wire-format change (`WriteBody.writeHistoryLink` is unchanged in shape).

## Recovery

`WriteSubtreeResolver::recover_wave` replaces `RotateScopeWritePlan::resume_write_scope_seed`, which was faked in every test. The production arm reads the scope pointer, verifies the owner-signed re-point object, requires `prevRoot` to be the root **this** rotation is moving off, runs the moved root through the full adoption gate, and opens its owner-write blob at the epoch the re-point names. The durable write-epoch floor is deliberately not raised here — a resume still re-reads the pre-wave root, whose write plane sits an epoch below.

`rotate_scope_write` then refuses:

- a recovered wave at a write epoch other than the one this run publishes at (the pointer plane has no adoption gate, so an older owner-signed re-point stays replayable at the scope's one stable pointer name);
- a recovered seed that does not derive the published root name — the reverse of the check `publish_moved` already makes forward.

A cut that does not advance the write epoch is refused release-active with `ResealError::HistoryLinkNotDescending`, never a `debug_assert!`; the ratchet only steps backward, so such a link is one every reader's walk drops. Covered by a test that also passes under `cargo test --release`.

## Tests

New, each proven red by reverting the production line via an edit and re-running:

- `a_write_cut_mints_its_history_link_over_the_retiring_seed`
- `a_write_cut_that_does_not_advance_the_write_epoch_is_refused_release_active`
- `a_planted_write_history_link_is_never_re_signed_into_the_moved_root`
- `the_recovery_seam_reads_the_fresh_seed_off_the_published_root`
- `a_scope_pointer_naming_another_predecessor_is_not_this_waves_recovery`
- `a_recovered_seed_that_does_not_derive_its_root_name_is_refused`
- `a_recovered_wave_at_another_write_epoch_is_refused_before_any_publish`
- `a_crash_before_the_root_publishes_has_no_seed_to_recover`

`mid_wave_crash_resumes_from_published_records_only` was rewritten to be honest: it now crashes at the canonical re-point (past the root republish, so published state holds the moved root), resumes with a fresh orchestrator on an entropy stream that would mint a *different* seed, and asserts the resume converged on the first run's published seed. The fake publisher records what the moved root's grant section publishes; the fake resolver reads it back. Nothing crosses the crash in memory.

## Gates

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, and `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown` all exit 0.

## Not fixed here

The non-atomicity between the write-floor guard and the root publish is untouched. Review findings on the `Carried` path's unauthenticated, unbounded bytes and on the minted link's audience are filed as follow-ups rather than folded in — they are pre-existing or spec-level, and nothing reads the field yet.

Closes #1139
Closes #1179

> [!NOTE]
> ### Fix write-plane history link minting and recover a resumed wave from published state
> - `rotate_scope_write` now calls `WriteSubtreeResolver::recover_wave` before minting fresh entropy; if a published seed is found, it validates the recovered epoch and root name before resuming, returning new non-retryable errors (`ResumedWaveAtAnotherEpoch`, `ResumedSeedNotAtItsRoot`) on mismatch.
> - `WriteWaveNet.recover_wave` implements recovery by reading the scope pointer and moved root to extract the write scope seed and epoch from the published owner-write blob.
> - `reseal_scope_root` now receives write-plane history context via `ResealSeeds.write_history` (either `WriteHistory::Carried` or `WriteHistory::Cut`) instead of `CommittedSet.write_history_link`, and mints a sealed history link during a write cut while rejecting non-advancing cuts with `ResealError::HistoryLinkNotDescending`.
> - `RotateScopeWritePlan` no longer accepts a caller-injected `resume_write_scope_seed`; resume seeds now come exclusively from `recover_wave` or fresh entropy.
> - Risk: `WriteSubtreeResolver` is a breaking trait change — all implementors must now provide `recover_wave`.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 28007a1.</sup>
> 
>

## Summary by CodeRabbit

* **Improvements**
  * Improved write-scope rotation and resealing by preserving or safely creating write history.
  * Enhanced crash recovery to restore write operations from published state with stronger epoch and root validation.
  * Added safeguards that reject stale or invalid history links and prevent unsafe re-signing.
* **Testing**
  * Expanded coverage for recovery scenarios, predecessor validation, seed derivation, and post-rotation safety.

### DIFF
--- a/crates/contract/tests/contract.rs
+++ b/crates/contract/tests/contract.rs
@@ -37,7 +37,7 @@ use cipherbox_engine::mailbox::poll_verified;
 use cipherbox_engine::net::REGISTRY_BATCH_MAX;
 use cipherbox_engine::rotation::{
     PrevEpochSeed, ResealSeeds, ResealedScopeRoot, ResolveFailure, ScopeRootIdentity,
-    ScopeRootPublishError, ScopeRootPublisher, SweepResolver, SweepTarget,
+    ScopeRootPublishError, ScopeRootPublisher, SweepResolver, SweepTarget, WriteHistory,
 };
 use cipherbox_engine::seams::{
     CredentialStore, Http, HttpCredentials, HttpMethod, HttpRequest, Mailbox,
@@ -1257,12 +1257,12 @@ async fn a_read_grant_delivers_its_share_pointer_through_the_live_mailbox() {
                 prev: None::<PrevEpochSeed<'_>>,
                 write_scope_seed: &parent_write_scope_seed,
                 write_epoch: 2,
+                write_history: WriteHistory::Carried(&[]),
                 pointer_read_key: &parent_pointer_read_key,
             },
             commitment: &parent_commitment,
             commitment_sig: &parent_commitment_sig,
             grant_ledger: &[],
-            write_history_link: &[],
             current_child_index: &[],
             carried_history_links: &[],
         },
--- a/crates/engine/examples/kat_gen.rs
+++ b/crates/engine/examples/kat_gen.rs
@@ -320,6 +320,7 @@ fn build_section_signer_vectors() -> (Vec<SectionSignerVector>, Vec<SectionSigne
         child_scope_index: Vec::new(),
         parent_node_seed: None,
         owner_write_blob_epoch: Some(OWNER_ROOT_EPOCH),
+        write_history_link: Vec::new(),
         grants: Vec::new(),
     });
     let owner_identity_pk = hex::encode(owner_identity.verifying_key().to_sec1());
--- a/crates/engine/src/facade.rs
+++ b/crates/engine/src/facade.rs
@@ -5085,6 +5085,7 @@ mod tests {
                 // cold-seeded write floor opens it and the owner recovers its
                 // write-scope seed for the held-set renewal signer.
                 owner_write_blob_epoch: Some(EPOCH),
+                write_history_link: Vec::new(),
                 grants: Vec::new(),
             });
             (fx.head_block, fx.head_cid_str, fx.name)
--- a/crates/engine/src/grants/create.rs
+++ b/crates/engine/src/grants/create.rs
@@ -46,8 +46,8 @@ use crate::grants::mint_grant_row;
 use crate::mailbox::post_sealed;
 use crate::rotation::{
     CommittedSet, ResealError, ResealSeeds, ResealedScopeRoot, ResolveFailure, ScopeRootIdentity,
-    ScopeRootPublishError, ScopeRootPublisher, SweepError, SweepResolver, derive_write_name,
-    reseal_scope_root, sweep_pass,
+    ScopeRootPublishError, ScopeRootPublisher, SweepError, SweepResolver, WriteHistory,
+    derive_write_name, reseal_scope_root, sweep_pass,
 };
 use crate::seams::{FloorStore, Mailbox, SeamError};
 use cipherbox_core::hex::lower as hex_lower;
@@ -133,8 +133,6 @@ pub struct ParentScopePlan<'a> {
     pub commitment_sig: &'a [u8; ECDSA_SIG_LEN],
     /// The parent's grant ledger (unchanged by this op).
     pub grant_ledger: &'a [GrantLedgerEntry],
-    /// The parent's write-plane history link (unchanged by this op).
-    pub write_history_link: &'a [u8],
     /// The parent's current direct-child-scope index (before this grant).
     pub current_child_index: &'a [ChildScopeRef],
     /// The parent's carried read-plane history links.
@@ -351,6 +349,7 @@ where
             prev: None,
             write_scope_seed: grantee.write_scope_seed,
             write_epoch: grantee.write_epoch,
+            write_history: WriteHistory::Carried(&[]),
             pointer_read_key: grantee.pointer_read_key,
         };
         // Mint-canonical: the adopted index carries the same canonicalization the
@@ -361,7 +360,6 @@ where
             commitment: &commitment,
             commitment_sig: &commitment_sig,
             grant_ledger: &ledger,
-            write_history_link: &[],
             direct_child_scope_index: &grantee_child_index,
         };
         reseal_scope_root(entropy, &identity, &seeds, &committed, &[])
@@ -414,6 +412,7 @@ where
             prev: None,
             write_scope_seed: &target.write_scope_seed,
             write_epoch: target.write_epoch,
+            write_history: WriteHistory::Carried(&target.write_history_link),
             pointer_read_key: &target.pointer_read_key,
         };
         let canonical_index = canonicalize(&target.direct_child_scope_index);
@@ -421,7 +420,6 @@ where
             commitment: &target.commitment,
             commitment_sig: &target.commitment_sig,
             grant_ledger: &target.grant_ledger,
-            write_history_link: &target.write_history_link,
             direct_child_scope_index: &canonical_index,
         };
         let section = reseal_scope_root(
@@ -466,7 +464,6 @@ where
             commitment: parent.commitment,
             commitment_sig: parent.commitment_sig,
             grant_ledger: parent.grant_ledger,
-            write_history_link: parent.write_history_link,
             direct_child_scope_index: &parent_index,
         };
         // The owner runs this leg, so the tag binding is not the caller's to
@@ -797,6 +794,7 @@ mod tests {
                 override_seed: &parent_override_seed,
                 read_epoch: 3,
                 prev: None::<PrevEpochSeed<'_>>,
+                write_history: WriteHistory::Carried(&[]),
                 write_scope_seed: &parent_write_scope_seed,
                 write_epoch: 2,
                 pointer_read_key: &parent_pointer_read_key,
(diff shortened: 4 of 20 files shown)


FSM1:
@coderabbitai pause

FSM1:
Both nitpicks from the 18:00Z review are addressed in `28007a12b`. Detail below, including one place where the fix went past what was asked and why.

## 1. Hoist the write-cut descending guard ahead of every seal

`crates/engine/src/rotation/reseal.rs`

The `WriteHistory::Cut` epoch guard ran after the grant blobs, owner blob, owner-write blob, ascent link and read-plane history link were already sealed. Every sibling release-active guard returns before any seal.

- `reseal.rs:374-379` — the guard now sits in the pre-seal block, immediately after the codec-bound guards and before `enforce_committed_ledger`.
- `reseal.rs:549-559` — the `WriteHistory::Cut` arm is a bare `mint_history_link(...)?`. Link generation is byte-identical.
- `reseal.rs:662-670` — `UndrawnEntropy`, the panic-on-draw entropy seam, moved from inside `a_committed_ledger_past_the_codec_bound_fails_closed_before_any_seal` to test-module scope, so both guards share one probe.
- `reseal.rs:1800` — the test is renamed `a_write_cut_that_does_not_advance_the_write_epoch_is_refused_before_any_seal`. It drives `UndrawnEntropy` and builds an interior scope root, so the ascent-link seal falls inside the pinned prefix too.

Red-proof: putting the guard back in the match arm makes that test fail with `panicked at reseal.rs:666: the guard must reject before any seal draws entropy`. It panics on the first grant-blob HPKE draw, which is direct evidence that the old placement sealed before refusing.

## 2. Terminal-verdict coverage — the list, not just the one variant

`crates/engine/src/net/rotation.rs:4325-4362`

The review asked for `ResealError::HistoryLinkNotDescending` to be added to the fixed terminal-variant list in `a_reseal_seam_failure_stays_retryable_and_a_trust_failure_does_not`. Doing only that would leave the actual defect in place: the list is hand-maintained, so any future variant can be omitted the same way and the test still passes.

Checking the list found **two** variants missing, not one — `HistoryLinkNotDescending` and `TagNotBoundToRecipient`. The second predates this PR.

So the array-of-terminals plus its separate entropy assertion is replaced by a single loop over one sample per variant, whose expected verdict comes from a `match` written independently of `reseal_verdict`.

**How far the compiler now enforces this.** The test's own `match` is exhaustive, so adding a `ResealError` variant fails to compile in the test as well as in production `reseal_verdict`. Verified: adding a probe variant and fixing only the production match still failed with `E0004: non-exhaustive patterns: &reseal::ResealError::ProbeVariant not covered --> crates/engine/src/net/rotation.rs:4344`.

**What it does not enforce.** Nothing makes the sample array itself hold every variant without a derive such as `strum::EnumIter`, which is not worth a new dependency for one test. The residual gap is narrow: an author adding a variant is stopped by the compiler and must write a classification arm, so omitting a sample becomes a visible choice rather than the silent omission this nitpick found.

Red-proof: re-classifying `HistoryLinkNotDescending` to `NotLanded` fails the test with `left: NotLanded, right: Rejected`.

## Gates

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` (806 passed) and `cargo check -p cipherbox-wasm --target wasm32-unknown-unknown` all exit 0. The diff is two Rust files and touches no TypeScript.

Both nitpicks were emitted in the review body rather than as inline threads, so there is nothing for `resolveReviewThread` to target and no thread was resolved.

Issue #1207, opened 2026-08-08
engine: the owner's contact book is plaintext at rest, disclosing the contact graph to a local reader


FSM1:
The owner's contact book (`crates/engine/src/grants/contact_store.rs`) writes det-CBOR in the clear under the host's `StagingStore`. Its neighbour under the same seam, the received-shares store, seals its body HPKE-to-self under the session `enc-subkey` before it reaches the host (`crates/engine/src/grants/received_share_store.rs`, `seal_received_shares`). The asymmetry is a privacy regression: a local reader of host storage learns the owner's whole contact graph, including every contact's identity key and X25519 encryption subkey.

This is a *new* disclosure rather than a restatement of a published one. The grant ledger that would otherwise carry those identity keys is sealed under the scope's write key, so before the contact book existed there was no plaintext-at-rest copy of the graph.

## What sealing also closes

The book carries no owner attestation and a contact code carries no counter. Sealing it HPKE-to-self puts the whole set under the owner's authority and closes three further residuals stated in the module header, none of which the per-entry binding verify can reach:

- a tamperer inserting any validly signed third-party code
- rolling an identity back to a subkey its holder signed once but has since rotated away from
- deleting an entry, so a later revoke reports `RecipientNotImported`

## Why it was not done in the PR that landed the book

Sealing needs a frozen structure tag of its own in `crates/core` — an HPKE info string, an AAD registry entry, a seal/open pair, and its KAT vectors. That is a change to the frozen core catalog, out of scope for an engine-layer store.

The invite-record store in #1165 is blocked on the same missing structure tag, for a stronger reason there (an unsealed record set is a forgery surface, not only a disclosure). Whoever lands the tag should land both consumers behind it.

## Acceptance

- A frozen structure tag for owner-local durable state, with KAT vectors, following the `received_shares` seal precedent.
- `StagingContactStore` seals its body before it reaches the host and opens on load; stored bytes that do not open fail closed, never read as an empty book.
- A test asserting no stored contact identity key appears in host bytes in the clear, mirroring `the_persisted_blob_never_holds_the_pointer_read_key_in_the_clear`.
- The plaintext-at-rest and no-owner-attestation residuals drop out of the `contact_store.rs` module header.

Follow-up from #1164, raised in review on #1196.

Issue #1165, opened 2026-08-07
engine: nothing persists a RecordedInvite, so a minted link cannot be converted or revoked in a later session


FSM1:
## The gap

`mint_invite_grant` (`crates/engine/src/grants/invite.rs:286`) returns a `MintedInvite` whose `link: RecordedInvite` (`crates/engine/src/grants/invite.rs:252`) is the owner's own authority over the link — "owner-local state, never network bytes" (`:242`), and the field doc says outright: "Persist it with the link" (`:272`).

Every later operation on a link takes that record back as input:

- `convert_invite_claim(.., links: &[RecordedInvite], ..)` — `crates/engine/src/grants/invite.rs:499`
- `revoke_invite_link(.., link: &RecordedInvite)` — `crates/engine/src/grants/invite.rs:639`

Nothing produces that slice outside tests. No store exists, and `SeamSet`'s nine fields (`crates/engine/src/seams/mod.rs:112`) have no durable home for it. So a link minted in one session cannot be converted or revoked in the next.

## Why the published row is not a substitute

The ledger row is deliberately byte-shaped like a personal grantee's, and the fields that would identify it as an invite — `recipientIdentityPk` and `expiresAt` — sit outside the owner's signature and are re-authorable by any write-grantee (`crates/engine/src/grants/invite.rs:242-251`). The `RecordedInvite.expires_at` copy is the authority; the published `expiresAt` is a cooperating-reader hint a write-grantee can strip or forge (`:259-262`, and the same rule at `:489`). An owner that reconstructs its links from the network is trusting exactly the field the design says not to trust.

## Checkable condition

- a link minted through `CreateInviteLink` (`crates/engine/src/facade.rs:491`) is recoverable after a session restart, and a claim arriving afterwards converts against the recovered record
- `revoke_invite_link` can be driven for a link minted in an earlier session
- a link whose recorded `expires_at` has passed is refused at conversion using the recorded copy, and a published `expiresAt` that disagrees does not change the verdict either way
- a mint whose record fails to persist does not return a usable URL fragment to the host — an unclaimable link is worse than a refused mint

## Scope note

#1128 needs owner-local state of the same shape for its already-converted claim ids and cut tags, and says so: "it belongs with whatever slice owns the owner's invite records". Land one store for the owner's invite state, not two.

Part of #635

## Shape — decided, do not re-litigate

**Do not add a tenth host seam.** Carry this store at the grants layer, over `StagingStore`, and
[...]

PR #1215, opened 2026-08-08
feat(core): seal owner-local state under one kind-separated structure

### LINKED ISSUE #1214: core: one sealed owner-local store primitive, kind-separated
## The gap

Three owner-local stores hold state the owner alone authors and reads, and each one currently answers the integrity question differently:

| Store | Today |
| --- | --- |
| received shares | sealed HPKE-to-self, `crates/core/src/seal/received_shares.rs` |
| contact book | **plaintext det-CBOR** under the same seam (#1207) |
| invite records | **does not exist yet** (#1165) |

The invite store is the one where unauthenticated bytes are an authority break, not only a disclosure.

`convert_invite_claim` matches a claim's sender against `link.ephemeral_identity_pk` (`crates/engine/src/grants/invite.rs:518`), re-derives the tag from `link.ephemeral_enc_pk` (`:527-531`), and reads the permission from the owner-signed commitment by that tag (`:535-541`). The tag binds the **encryption** half only, so `ephemeral_identity_pk` is bound by nothing in the record.

A party who can write the store therefore pairs a real link's `ephemeral_enc_pk` — which re-derives to a tag matching a real committed entry, carrying its real permission — with an identity key it holds. A claim signed by that key converts, and the owner mints a genuine grant at up to `Write` with no owner key material involved. `RecordedInvite.expires_at` is likewise the authority (`invite.rs:259-262`), so an expired link is resurrected the same way.

## Decision

**One generic owner-local sealed store in `crates/core`, with the store kind bound into the HPKE `info`**, replacing the per-store approach. It serves invite records, the contact book, and received shares.

Rationale: three stores is enough repetition to justify one primitive, and one KAT set governing every owner-local store beats three that drift. The alternative — a third module copied from `received_shares.rs` — keeps domain separation correct by construction but multiplies the fail-closed load logic that #1185 and #1196 already had to get right once each.

**Migrating `received_shares` onto it is free today and stops being free at the cutover.** v2 has never been deployed; `blueprint/deploy.md` makes the staging cutover a redeploy rather than a migration (FSM1/cipher-box-next#6 removes all data-continuity obligations). Doing this after #653 means a format migration instead of a redefinition.

## Scope

1. A decision record in `FSM1/cipher-box-next` — this defines a frozen structure and supersedes the shape `received_shares.rs` established, so it belongs in the corpus rather than only in a PR body.
2. The `blueprint/core.md` wire-format entry for the new structure, and removal of the per-store entry it replaces.
3. The core primitive: header, version, kind discriminator, seal and open, over `hpke_seal_auth` as `received_shares.rs` does today.
4. **KAT vectors, including a cross-kind negative**: a blob sealed under one kind must fail to open under another. The discriminator is where this design can go wrong, so the KAT is the guardrail, not a nicety.
5. Migrate `received_shares` onto the primitive and delete the superseded module.
6. Encode/decode fail-closed symmetry: every invariant the open path hard-rejects gets a release-active check on the seal path returning `Err`, never `debug_assert!`.

## Not in scope

The engine-side stores themselves. #1165 builds the invite store over this primitive; #1207 seals the contact book with it. Both are blocked by this issue.

## Checkable condition

- a blob sealed as one store kind fails to open as another, proven by a KAT vector rather than a unit test alone
- `received_shares` round-trips through the new primitive with its old module gone
- a tampered blob fails to open rather than opening to altered content

### LINKED ISSUE #1232: test(core): the enc-subkey info-distinctness assertion fails open when a structure is added
## The defect

The enc-subkey `info`-distinctness assertion in `crates/core/tests/kat_manifest.rs`
(around `:4853-4857`) **fails open** at exactly the moment it is supposed to fire.

It compares the deduplicated set size against a computed `3 + m.owner_local.kinds.len()`
rather than against `infos.len()`. Today both are 6, so it passes honestly. The
hole opens when someone adds a fourth enc-subkey structure and pushes its `info`
literal onto the vec:

- the new literal is **distinct** → set size 7, expected 6 → the test fails, the
  author bumps the `3`, no harm done
- the new literal **collides** with an existing one → set size 6, expected 6 →
  **the test passes**, and two structures now share a key schedule

Two structures sharing an HPKE `info` share a key schedule, which makes their
blobs transplantable between families. This assertion is the only *general* guard
against that. The pairwise transplant tests — `an_op_record_and_a_content_key_blob_never_open_as_each_other`
and its siblings — only cover pairs someone already thought to write, so they
cannot catch a newly added structure colliding with an existing one.

## The fix

Compare against `infos.len()`, which has no such hole. Two lines: swap the
expected count and drop the now-unused `let m = manifest();`. Test-only, no vector
regeneration, no production logic.

## Why it was not fixed in the PR that introduced it

Found as a CodeRabbit nitpick on FSM1/cipher-box#1215, whose verdict was otherwise
nitpick-only with clean CI. Pushing to that ready PR would have started an
incremental review and consumed a slot from a ceiling of one review per hour,
which several PRs were queued behind.

Deferring costs no exposure: the only way to trip the fail-open is a PR that adds
a new enc-subkey `info` literal, and such a PR necessarily edits this same test.

## Fold into

The next PR touching `crates/core/tests/kat_manifest.rs` or `crates/core/src/seal/`.

While there, tighten `seal_owner_local`'s ephemeral doc at
`crates/core/src/seal/owner_local.rs:123-125`. It says reuse "under one recipient
key" is a break, dropping the `info` qualifier that `crates/core/src/suite/hpke.rs:74`
carries. The stricter phrasing is wrong: the break needs the same ephemeral,
sender, recipient **and** `info`.

FSM1:
## What

One generic owner-local sealed structure in `crates/core/src/seal/owner_local.rs`, replacing the per-store approach, and `received_shares` migrated onto it with its module deleted.

Implements [ADR 0006](https://github.com/FSM1/cipher-box-next/blob/main/decisions/0006-owner-local-sealed-store.md) (FSM1/cipher-box-next#60).

## The structure

`seal_owner_local(owner_enc_secret, kind, ephemeral_scalar, body)` / `open_owner_local(owner_enc_secret, kind, blob)`, HPKE **auth mode to self** under the owner's enc subkey, structure tag `owner-local` (`0x0d`).

The stored frame is the settings record's three clear keys — `v`, `enc`, `ciphertext`. Neither the owner tag nor the kind is serialized:

| Input | Where it binds |
| --- | --- |
| `version` | clear header **and** AAD — rewriting the clear copy fails the tag |
| `kind` name | HPKE `info` = `cipherbox/v2/owner-local/<name>` |
| `kind` discriminator | AAD element 4 |
| `owner_tag` | AAD element 5, rebuilt from the opening key |

**The kind is a key-schedule input, never a wire field.** The reader always knows which store it opened, so there is nothing to compare and nothing to relabel: a blob offered as the wrong store is refused by the AEAD. That is what makes cross-kind separation a decryption failure rather than a parse failure (ADR 0006 consequence 1), and it is why the kind on the wire would be a downgrade rather than a diagnostic.

Kinds are a frozen registry — `received-shares` (`0x01`), `contact-book` (`0x02`), `invite-records` (`0x03`) — pinned name-for-byte-for-info-string in `kat/manifest.json`, with injectivity asserted on both sides.

## Cross-kind KAT

`vectors/owner_local/owner_local_reject.json` carries a negative for **every ordered pair** of kinds (6 of them). `owner_local_cross_kind_vectors_cover_every_ordered_pair` asserts each one fails `hpke-open-failed` / `trust` — a decryption failure, not a parse failure — *and* that the same blob still opens under the kind it was sealed as, so the pair proves separation rather than a broken blob.

Vectors regenerated through `cargo run -p cipherbox-core --example kat_gen`; none hand-edited.

## The migration

`crates/engine/src/grants/received_share_store.rs` names `OwnerLocalKind::ReceivedShares` on both sides. v2 has never been deployed, so this is a redefinition, not a data migration — no shim, and tag `0x0d` is redefined from `received-shares` to `owner-local`.

The fail-closed load rule is unchanged and now has a third arm: a sibling store's blob whose body is a list this build decodes perfectly still surfaces as an error, never as a list to adopt.

## Fail-closed symmetry

The open path hard-rejects a forward `v`, an unknown clear-header field, a missing frame field, a short/low-order `enc`, and any AEAD failure. None of those is reachable from the seal path's inputs: the version and the frame key set are constants and `enc` is HPKE's own output, so there is no encode-side invariant left to guard with a runtime check. What could have been violated is closed by typing instead — `owner_tag` is `[u8; 32]` rather than a `Vec<u8>`, and `kind` is an enum — and the emitted key set is asserted equal to the reader's exhaustive `FRAME_KEYS`.

## Tests

Every new assertion was red-proved by breaking the code it guards:

| Break | Goes red |
| --- | --- |
| kind dropped from both the `info` and the AAD | `a_blob_sealed_under_one_kind_never_opens_under_another`, `owner_local_cross_kind_vectors_cover_every_ordered_pair`, `owner_local_accept_vectors_seal_reproduce_and_open`, `a_blob_from_another_owner_local_store_fails_closed` |
| two kinds share a name | `the_kind_registry_is_injective`, `owner_local_kind_registry_is_frozen` |
| kind added as a fourth clear frame key | `the_stored_blob_names_no_key_and_no_kind`, `owner_local_accept_vectors_seal_reproduce_and_open` |
| a kind's accept vector dropped from the generator | `owner_local_accept_vectors_seal_reproduce_and_open` |

All in the **Core KATs** and **Test** gates.

## Review gates

`/simplify`, `/security-review`, and `/crypto-privacy-review` all run (the diff touches `crates/core` seal material). No CRITICAL or HIGH findings. Folded in:

- **`ALL` completeness was a process promise, not a check.** Every separation proof iterates `OwnerLocalKind::ALL`; nothing forced a new variant into it, so a variant with a duplicate name and discriminator could have unified with an existing store on a green build. `the_kind_registry_lists_every_kind_exactly_once` now routes each kind through an exhaustive `match` — a fourth variant is a **compile error** until it is registered.
- **The structure buys no freshness**, and the `invite-records` kind is where that matters (conversion reads a link's permission and deadline from it). The module header now says so, and says one kind is one blob per owner because the AAD binds no instance id.
- **`contact_store.rs`'s module doc asserted a blocker this PR removes** ("needs a frozen structure tag of its own"). Rewritten to name what sealing does and does not close.
- **Cross-family `info` uniqueness is now tested**, not merely true by inspection — `owner-local` is the first family whose `info` string is computed rather than a literal, which is where a collision first becomes possible.
- **The cross-kind KAT probes get distinct ephemerals.** Sharing one was safe only because the kind is in the `info`; a corpus should not model a pattern whose safety is the thing under test.
- Comment/test trims: the tautological `starts_with` assert, the injectivity duplication between the unit test and the manifest test, and three restatements of the cross-kind rationale.

Pushed back on, with reasons:

- **Bumping `OWNER_LOCAL_V` to 2** so pre-PR received-shares bytes classify `unsupported-record-version` instead of `hpke-open-failed`. The engine errors on `load()` either way — the class does not un-wedge anything — and a brand-new structure starting at v2 to be readable-as-legacy is the compatibility affordance this slice was told not to add. v2 has never been deployed.
- **Dropping the unused `ContactBook` / `InviteRecords` kinds.** They are the two consumers this slice unblocks (#1207, #1165), and the ADR names all three. Freezing two kinds later means re-freezing the KAT.
- **Extracting the shared `{v, enc, ciphertext}` frame across `settings_record` / `content_key` / `op_record`.** Real duplication, but a four-module refactor of frozen formats does not belong in the PR that redefines a structure tag.
- **Adding a generation counter to the AAD.** Needs an engine-held high-water mark; that is the invite-store slice's call, and the limitation is now stated at the type instead.
- **Binding the kind to the staging-key prefix.** Engine-side, and the other two stores do not exist yet.

## Not in scope

The engine-side stores. #1165 builds the invite store over this primitive and #1207 seals the contact book with it; both are unblocked by this.

Closes #1214
Closes #1232

> [!NOTE]
> ### Unify owner-local sealed stores under one kind-separated structure replacing `received-shares`
> - Introduces a new `owner_local` module in [owner_local.rs](https://github.com/FSM1/cipher-box/pull/1215/files#diff-5acbaf4bdd4c295d6de600f2d4010c26c93cbed8b258b29e6d5eb7241749ce5f) with a generic `OwnerLocalKind` enum (`ReceivedShares`, `ContactBook`, `InviteRecords`), each with its own HPKE info string and discriminator, all sharing struct tag `0x0d`.
> - Replaces the old `received-shares`-specific seal/open API (`seal_received_shares`, `open_received_shares`) with `seal_owner_local`/`open_owner_local` that take a kind parameter; cross-kind opens fail at decryption.
> - Updates `received_share_store` and `contact_store` in the engine to use the new unified API with `OwnerLocalKind::ReceivedShares` and `OwnerLocalKind::ContactBook` respectively.
> - KAT vectors and manifest are regenerated: accept vectors cover each kind with empty and populated bodies; reject vectors add exhaustive cross-kind negative cases for every ordered pair of distinct kinds.
> - Behavioral Change: blobs sealed under one owner-local kind (e.g. `ContactBook`) now fail to open as another kind (e.g. `ReceivedShares`), enforced via AAD and HPKE info binding.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 16fc5f0.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Added encrypted owner-local storage supporting received shares, contact-book data, and invite records.
  - Added authenticated, owner-bound protection with validation against tampering, malformed data, and cross-type reuse.
  - Preserved secure received-share storage through the new unified format.

- **Bug Fixes**
  - Improved isolation between stored data types and owners, preventing incompatible or forged records from being accepted.

- **Tests**
  - Added comprehensive acceptance and rejection coverage for valid, tampered, malformed, and cross-type records.


FSM1:
@coderabbitai — one of the three is a real fail-open and is fixed; the other two are declined, with the mechanism for each.

## Fixed: `kat_manifest.rs` distinct-info count

Correct, and worse than a maintainability nit — the assertion **failed open**. Comparing the deduplicated set size against a recomputed `3 + m.owner_local.kinds.len()` means a future structure whose `info` literal *collides* with an existing one leaves both sides equal and the test green: set size 6, expected 6. Two structures would then share a key schedule and their blobs would be transplantable between families, which is exactly what this test exists to prevent — the pairwise transplant tests only cover pairs someone already wrote.

Red-proved: with a fourth entry added to `infos` whose literal duplicates `OP_RECORD_HPKE_INFO`, the old assertion **passes** (6 == 3 + 3) and the `infos.len()` form **fails** (6 != 7). Fixed, `m` dropped, KAT corpus byte-identical after regeneration.

## Declined: isolating the AAD discriminator from the HPKE `info`

The proposed test cannot be written against the public API. `seal_owner_local` and `open_owner_local` both derive `info` from the `kind` argument (`owner_local.rs:142`, `:195`), so "one fixed `info` for every kind, per-kind AAD" would have to hand-roll `hpke::hpke_seal_auth` in a configuration no production path can construct — the test would assert a property of a code path that does not exist.

What it would assert is already covered generically at `suite/hpke.rs:555-565` ("Wrong AAD") and `:566-575` ("Wrong info"): AAD and `info` each independently refuse an open. Nothing about owner-local changes that.

The risk named — a future change collapsing the per-kind `info` to one family string — cannot land silently. `kat/manifest.json` freezes all three `hpkeInfo` values, and `owner_local_kind_registry_is_frozen` asserts `spec.hpke_info == kind.hpke_info()` per kind (`kat_manifest.rs:4830-4835`). That collapse turns the Core KATs gate red.

## Declined: distinct ephemeral for the cross-family transplant

No key or nonce reuse occurs. The reuse invariant is scoped to one recipient **and** `info` (`suite/hpke.rs:70-79`: "the same `recipient_pub`/`info`"). The two seals use `cipherbox/v2/owner-local/received-shares` and `cipherbox/v2/settings-record`, so the key schedules diverge and the derived key and base nonce differ.

The shared ephemeral is the *sharper* vector. Both seals are HPKE auth mode to the same owner key, so one ephemeral forces an identical KEM shared secret — which makes the transplant's rejection attributable to the `info` and the struct tag alone, and nothing else. A distinct ephemeral would let a reader wonder whether the KEM did the work.

This is the established convention, not a new one: `kat_manifest.rs:4010-4022` on `main` seals an op record and a content-key blob under one ephemeral precisely so they "differ only in their HPKE `info` and AAD".

On the apparent tension with the cross-kind comment at `kat_gen.rs:7080-7083` — the distinction is real. Owner-local `info` strings are *computed* (`prefix + name`), so a change that collapses them to the prefix is representable in one edit; the cross-kind probes take distinct ephemerals so the corpus does not lean on that computation. Cross-family `info` strings are separate frozen literals in different families; there is no edit that makes `cipherbox/v2/settings-record` equal an owner-local `info`.

Changing it would also regenerate frozen corpus bytes and leave the repo holding two contradictory conventions for the same situation.

Issue #1253, opened 2026-08-11
identity: implement ADR 0008 — CipherBox issues the Core Kit identity token


FSM1:
Umbrella for implementing **ADR 0008** — CipherBox issues the Core Kit identity token, and wallet login is first-class on web (FSM1/cipher-box-next#64). The blueprint statement of it is #1255.

Work is tracked as sub-issues of this one — this body carries no checkbox list.

## Correction to this issue's original framing

This issue was filed asserting that v2 had regressed from v1 by accident. That is wrong, and the correction matters for scoping.

A CipherBox login has **two** authentications:

1. **Unlocking the Core Kit** — an identity provider vouches for the person; the Core Kit returns the TSS key the login secret derives from.
2. **Authenticating to the API** — the engine signs a challenge with the derived identity key.

(2) was decided — FSM1/cipher-box-next#28 D5, restated in `blueprint/engine.md` and `blueprint/api.md`. It is correct and nothing here changes it.

(1) was never decided. v2 built it as Web3Auth's hosted OAuth in the client wiring; v1 had CipherBox mint a JWT against a custom verifier. The account model cannot distinguish them, because the Core Kit yields the same secp256k1 key whichever provider vouched — so `blueprint/api.md`'s "Account = the Web3Auth-derived identity key" holds under both, and the issuer was chosen rather than decided.

ADR 0008 decides it. That makes the change far smaller than "restore v1": the account model, challenge-signature login, the engine-held token lifecycle, and the `CredentialStore` split all stand.

## What the ADR decides

- **D1** — CipherBox issues the identity token the Core Kit consumes, via a custom verifier over the API's own JWKS. Passwordless email returns to the API. Google's OAuth client ID is the provider's, distinct from the Web3Auth project client ID.
- **D2** — Wallet is a first-class **first** login, **web only**. This amends FSM1/cipher-box-next#5. It is web-only because the Tauri webview reaches no wallet — a platform property, not a deferred feature. v1 wrote that desktop path and never wired it.
- **D3** — One host-agnostic orchestration package; credential collection injected per host. The boundary sits at credential collection, one step earlier than v1 drew it at the bearer token, which is where v1's two hosts drifted.

## Why the boundary moved

[...]

PR #1272, opened 2026-08-11
fix: carry the Google and Web3Auth client IDs separately

### LINKED ISSUE #1256: fix: Google sign-in sends the Web3Auth client ID to Google
Google sign-in fails with `401 invalid_client` — "The OAuth client was not found" — for every account, because the app sends Google the Web3Auth **project** client ID where Google's own OAuth client ID belongs.

Independent of ADR 0008: this is wrong under either identity architecture, and it should land first.

## Mechanism

`apps/web/src/auth/coreKit.ts` sources one value and spends it on two unrelated things:

```
:141  web3AuthClientId: clientId,                    // correct — the Web3Auth project
:150  new Web3AuthSession(coreKit, store, verifier, clientId)
:62     clientId: this.clientId,                     // wrong — this reaches Google
```

`loginEnv` (`apps/web/src/engine/config.ts:104`) reads that single value from `VITE_WEB3AUTH_CLIENT_ID`. Inside `subVerifierDetails`, `clientId` is the **provider's** client ID — for `typeOfLogin: 'google'`, a Google Cloud OAuth 2.0 ID registered against the verifier. The two identifiers do not share a format, so Google rejects the request before Web3Auth is involved.

## It is not only local

`.github/workflows/deploy-staging.yml` still passes `VITE_GOOGLE_CLIENT_ID: ${{ vars.GOOGLE_CLIENT_ID }}` at `:101`, `:179`, `:273` and `:349`, and **no v2 source reads it** — absent from `apps/web/src`, undeclared in `apps/web/src/vite-env.d.ts`. The deployment supplies the right value; the code no longer collects it. Staging Google sign-in should be failing identically, untested.

`v1` read it at `apps/web/src/components/auth/GoogleLoginButton.tsx:53`. The variable survived the rewrite; its reader did not.

## Work

- Carry the two client IDs separately: `VITE_WEB3AUTH_CLIENT_ID` for `Web3AuthMPCCoreKit`, the provider's for `subVerifierDetails`.
- Restore `VITE_GOOGLE_CLIENT_ID` as its reader and declare it in `vite-env.d.ts`.
- Add it to `apps/web/.env.example`, which currently implies one client ID is enough — that is how this recurs.
- Check `email_passwordless`: it takes the same `subVerifierDetails.clientId`.

## Acceptance

- Google sign-in completes against a verifier configured with a Google OAuth client ID.
- A missing provider client ID fails naming which one is absent, rather than surfacing a provider error the member cannot act on.
- No deployment sets a variable nothing reads.

Part of #1253

FSM1:
Closes #1256. Part of #1253.

## What was wrong

`subVerifierDetails.clientId` is the **provider's** client ID. The app was spending one value on two unrelated things: `VITE_WEB3AUTH_CLIENT_ID` correctly named the Web3Auth project to the `Web3AuthMPCCoreKit` constructor, and then that same string was handed to the Google sub-verifier. Google has never issued it, so every Google sign-in ended at `401 invalid_client` before Web3Auth was involved.

The right value was never absent from the deployment — `VITE_GOOGLE_CLIENT_ID` is passed at every staging build site in `deploy-staging.yml` and `desktop-staging-release.yml`. It survived the rewrite; its reader did not.

## What changed

- `loginEnv` now returns `{ web3AuthClientId, googleClientId, verifier }`. The old shape returned `clientId`, a name that invited exactly this substitution — neither field can now be picked by accident.
- `VITE_GOOGLE_CLIENT_ID` joins `LOGIN_ENV`, so it is declared in `vite-env.d.ts`, named by `missingLoginEnv`, and — through `DEPLOY_ENV` — fails the `deployEnvGate` build plugin on a staging or production build that omits it. A deployed build can no longer ship a Google button that cannot work.
- The session carries a per-method map of client IDs instead of one string, so each sub-verifier is routed to its own registration at the call site.
- `apps/web/.env.example` now lists all three and states plainly that the two client IDs are separate registrations and not interchangeable — the template previously implied one client ID was the whole story, which is how this recurs.

## `email_passwordless`

It takes the **Web3Auth project** client ID, not Google's. That sub-verifier is Torus-hosted: there is no third-party OAuth client behind it, and the connection is registered under the Web3Auth project itself. Sending Google's ID there would break email sign-in in the same way this PR fixes Google. Worth a second pair of eyes against the verifier's actual dashboard configuration before merge.

## Tests

- `coreKit.test.ts` captures `subVerifierDetails` off the substituted SDK and asserts the Google connection receives the Google OAuth client ID, the email connection receives the Web3Auth one, and the constructor receives the Web3Auth one.
- `config.test.ts` asserts the two IDs stay apart through `loginEnv`, and that a build missing `VITE_GOOGLE_CLIENT_ID` throws naming it — absent, blank, and whitespace-only alike — rather than surfacing a provider error nobody can act on. `missingDeployEnv` covers the red-build half.
- Mutation-checked: restoring the old behaviour — routing Google to the Web3Auth ID — fails the Google routing test and nothing else.

Ran: `pnpm --filter @cipherbox/web test` (32 files, 295 tests, includes `tsc -b`), `pnpm lint`, `pnpm lint:md`, `pnpm lint:tracker-refs` — all green.

## Needs human verification

No dev server or configured verifier exists in this environment, so the runtime path is unverified: **a real Google sign-in against a verifier whose Google sub-verifier is registered with the Google Cloud OAuth client ID in `GOOGLE_CLIENT_ID`**, plus one email sign-in to confirm that connection still authenticates on the Web3Auth project ID.

## Scope

Only the client-ID plumbing. The Core Kit login architecture is untouched — `loginWithOAuth` is replaced wholesale by a separate PR.

> [!NOTE]
> ### Carry Google and Web3Auth client IDs separately for OAuth sub-verifiers
> - Splits the single `clientId` into two: `VITE_GOOGLE_CLIENT_ID` (used by Google OAuth sub-verifier) and `VITE_WEB3AUTH_CLIENT_ID` (used by the email passwordless sub-verifier and SDK init). Previously both sub-verifiers shared one client ID.
> - Updates [`config.ts`](https://github.com/FSM1/cipher-box/pull/1272/files#diff-4e12e157a8de69cb09afa15f6591facf5c9392fe6a014bd8f63a3804d4ecfaad) to require and return both IDs from `loginEnv()`, throwing if either is missing or blank.
> - Updates [`coreKit.ts`](https://github.com/FSM1/cipher-box/pull/1272/files#diff-ef6c9dcd7f99504fee0b68ee69636759b18751c9bbb3433d427c06efce875ec0) to accept a `SubVerifierClientIds` map and select the correct ID per login method.
> - Behavioral Change: `VITE_GOOGLE_CLIENT_ID` is now a required environment variable; deployments missing it will throw at session creation time.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9f87e8d. (Automatic summaries will resume when PR exits draft mode or review begins).</sup>
> 
>

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,9 +181,9 @@ the same backend, or the republisher re-PUTs into a store the client never reads
 The API's write path is live end to end: authenticate and `POST /content/upload`
 returns 201 with bytes pinned in the local Kubo.
 
-Interactive login through the web UI needs `VITE_WEB3AUTH_CLIENT_ID` and
-`VITE_WEB3AUTH_VERIFIER`, which a clean checkout does not carry — the UI boots and
-renders without them, but a Core Kit session cannot be created. The suites that need an
+Interactive login through the web UI needs `VITE_WEB3AUTH_CLIENT_ID`,
+`VITE_WEB3AUTH_VERIFIER` and `VITE_GOOGLE_CLIENT_ID`, which a clean checkout does not
+carry — the UI boots and renders without them, but a Core Kit session cannot be created. The suites that need an
 authenticated session use the build-time introspection hook instead; see
 [`tests/web-e2e/README.md`](tests/web-e2e/README.md).
 
--- a/README.md
+++ b/README.md
@@ -183,9 +183,9 @@ returns 201 with bytes pinned in the local Kubo.
 
 Interactive login through the web UI needs `VITE_WEB3AUTH_CLIENT_ID`,
 `VITE_WEB3AUTH_VERIFIER` and `VITE_GOOGLE_CLIENT_ID`, which a clean checkout does not
-carry — the UI boots and renders without them, but a Core Kit session cannot be created. The suites that need an
-authenticated session use the build-time introspection hook instead; see
-[`tests/web-e2e/README.md`](tests/web-e2e/README.md).
+carry — the UI boots and renders without them, but a Core Kit session cannot be created.
+The suites that need an authenticated session use the build-time introspection hook
+instead; see [`tests/web-e2e/README.md`](tests/web-e2e/README.md).
 
 A first folder create does not yet publish, because nothing provisions a fresh account's
 first vault pointer, so its writes are accepted, rendered pending, and reach no endpoint.
--- a/apps/web/.env.example
+++ b/apps/web/.env.example
@@ -22,7 +22,14 @@ VITE_ROUTING_ENDPOINTS=http://localhost:3001
 # Optional comma-separated public gateways the engine may read from.
 # VITE_PUBLIC_GATEWAYS=
 
-# Web3Auth Core Kit. Interactive login needs both. A clean checkout carries
-# neither, so the UI renders but cannot create a session until they are set.
+# Web3Auth Core Kit. Interactive login needs all three, and a clean checkout
+# carries none — the UI renders but cannot create a session until they are set.
+#
+# The two client IDs are separate registrations and are NOT interchangeable:
+#   VITE_WEB3AUTH_CLIENT_ID — the Web3Auth project, from the Web3Auth dashboard
+#   VITE_GOOGLE_CLIENT_ID   — the Google Cloud OAuth 2.0 client the verifier's
+#                             Google sub-verifier is registered against
+# Sending the Web3Auth one to Google is a 401 invalid_client on every account.
 # VITE_WEB3AUTH_CLIENT_ID=
 # VITE_WEB3AUTH_VERIFIER=
+# VITE_GOOGLE_CLIENT_ID=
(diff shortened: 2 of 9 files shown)


FSM1:
Closing unmerged — superseded by #1273, which carries both of these commits and now closes #1256 itself.

The env plumbing here is load-bearing and survives untouched: `VITE_GOOGLE_CLIENT_ID` declared, in `LOGIN_ENV` and so the deploy gate, returned by `loginEnv` — #1273 consumes it for Google credential collection via `main.tsx` → `IdentityProvider` → `LoginPage`. What does not survive is the `subVerifierDetails` routing, because #1273 replaces `loginWithOAuth` with `loginWithJWT`, which has no sub-verifier details at all.

With no staging environment yet, landing an intermediate fix to a code path the next PR deletes bought nothing, and the surviving half reaches `main` either way. Reviewing it once, in #1273, next to the code that actually uses it, is the better trade.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1273, opened 2026-08-11
feat: CipherBox issues the identity token, and wallet login is a first login

### LINKED ISSUE #1256: fix: Google sign-in sends the Web3Auth client ID to Google

### LINKED ISSUE #1257: api: issue the identity token the Core Kit consumes
Make the API the issuer of the identity token the Core Kit consumes — ADR 0008 D1.

Today the Core Kit authenticates against Web3Auth's hosted OAuth (`apps/web/src/auth/coreKit.ts:59` calls `loginWithOAuth`), so a method that authenticates against CipherBox first has no way to produce a login secret.

## Work

- A JWKS endpoint the Web3Auth custom verifier can fetch.
- A CipherBox JWT minted per verified method, with the claims the verifier is configured to expect.
- Extend the wallet path: `apps/api/src/auth/services/siwe.service.ts` already verifies a signature; it must mint the identity token for the `login` intent. v1's contract carried both intents — `login` "creates/finds a user", `link` "only verifies ownership".
- The client switches `loginWithOAuth` to `loginWithJWT` against the CipherBox verifier.

## Constraints

The account model does not change: the Core Kit yields the same secp256k1 identity key whichever provider vouched, and challenge-signature login to the API (FSM1/cipher-box-next#28 D5) is untouched. This changes **which provider unlocks the Core Kit**, nothing downstream of it.

A JWKS caching hazard is already on record from v1, in exactly this handoff — read `.learnings/2026-02-16-ephemeral-jwks-web3auth-caching.md` on branch `v1` before writing the verification path.

## Acceptance

- The Core Kit logs in against a CipherBox-issued token verified over CipherBox's JWKS.
- A token signed by anything else is refused.
- The derived identity key is the same one challenge-signature login already authenticates.

Part of #1253

### LINKED ISSUE #1258: api: own passwordless email again
Return passwordless email to CipherBox — ADR 0008 D1.

v2 delegates it to Web3Auth's `email_passwordless` sub-verifier, so code delivery, deliverability, rate limits and branding sit outside CipherBox. v1 owned it: `apps/api/src/auth/services/email-otp.service.ts`, with `sendOtpDto` / `verifyOtpDto`.

## Work

- Reinstate the OTP send/verify pair on the API, with a configured delivery provider rather than an assumed one.
- On verify, mint the identity token (the surface from the API-issuer slice).
- Replace the client's `email_passwordless` sub-verifier path with the CipherBox flow.

## Acceptance

- A member receives a code from CipherBox and signs in with it.
- A code CipherBox did not issue is refused, as is an expired one.
- The delivery provider is configuration, and its absence fails loudly at start rather than at first send.

Part of #1253

### LINKED ISSUE #1260: web: make wallet a first-class first login
Make wallet a first-class first login on web — ADR 0008 D2, which amends FSM1/cipher-box-next#5.

Today `WalletLoginButton` renders on the cold login page and every click dead-ends: `Command::SiweLogin` is refused with `NotStarted` before `facade.start`, and the member sees "engine not started" — internal state, not anything actionable.

Once the API issues the identity token, a wallet signature reaches the same derived key as any other method, so no separate account model is needed.

## Work

- wagmi collects the signature on the UI thread, as it does now.
- The API verifies it and mints the identity token (the API-issuer slice).
- The Core Kit login proceeds identically to Google and email; `start(secret)` follows.
- The affordance stops dead-ending, and its failure messages describe what the member can do.

## Scope

Web only. The Tauri webview reaches no wallet, so desktop offers the method set minus wallet — absent rather than present and unable to complete. v1 wrote a desktop `loginWithWallet` and never wired it: no UI trigger, never exercised.

## Consequences for #938

#938 proposes moving the wallet affordance off the login page onto a started-engine surface. That accommodated a limitation this removes, so its premise goes — but the linked-methods surface it describes is still worth building for **linking** a wallet to an existing account, which is a separate intent from logging in with one.

## Acceptance

- A member with no prior session signs in with a wallet and reaches a working vault.
- The account is the same one the other methods reach.
- No affordance on the login page can be clicked into a dead end.

Part of #1253

FSM1:
Implements [ADR 0008](https://github.com/FSM1/cipher-box-next/blob/main/decisions/0008-cipherbox-issues-the-identity-token.md) D1 and D2.

Closes #1256. Closes #1257. Closes #1258. Closes #1260. Part of #1253.

Supersedes #1272, which is closed unmerged — its two commits are carried here. There is no staging environment yet, so landing an intermediate fix to a code path this PR deletes bought nothing.

## Why these ship together

The Core Kit derives its TSS key from the `(verifier, verifierId)` pair. Moving from Web3Auth's hosted verifier to a CipherBox custom verifier changes the derived key. If Google moved to the CipherBox verifier while email stayed on the `email_passwordless` sub-verifier, one person would land on two different accounts depending on which button they pressed. All three methods have to cross in one commit, so they do.

Identities created under the old Web3Auth-hosted verifier are not reachable after this change. That is fine because no accounts exist yet; there is deliberately no migration, dual-verifier fallback, or compatibility shim.

## What changes

**The API issues the identity token (#1257).** A JWKS endpoint plus an RS256 mint. Each verified method produces a CipherBox JWT, and the Core Kit redeems it through `loginWithJWT` against CipherBox's own JWKS instead of `loginWithOAuth`.

**The API owns passwordless email again (#1258).** It issues the code, verifies it, and owns delivery behind a provider seam. The code is now collected in the app rather than in a provider window.

**Wallet is a first-class first login, web only (#1260).** `WalletLoginButton` no longer routes to the engine's `Command::SiweLogin`, which was refused with `NotStarted` and surfaced to the member as "engine not started". A SIWE signature now mints the same token as any other method and lands on the same Core Kit login and secret handoff, so a member with no prior session reaches a working vault.

**The two client IDs stop being conflated (#1256).** `subVerifierDetails.clientId` is the provider's client ID, but the app was passing the Web3Auth project client ID, so every Google sign-in ended at `401 invalid_client` before Web3Auth was involved. `VITE_GOOGLE_CLIENT_ID` is declared, joins `LOGIN_ENV` and so the deploy gate, and `loginEnv` returns it — which the Google credential collection here consumes via `main.tsx` → `IdentityProvider` → `LoginPage`. The per-method sub-verifier client-ID map from that fix is deleted by this one: `loginWithJWT` takes no `subVerifierDetails`.

## The subject mapping

A verified provider identity maps to a stable subject id through a new `identity_subjects` table. That id is the JWT `sub` and the `verifierId` passed to `loginWithJWT`.

It carries **no `user_id`**, deliberately. The account still materializes at `POST /auth/login` keyed by the derived `publicKey`, exactly as before — this table's only job is to yield a stable `verifierId`, so it cannot fork the account model. Keeping `user_id` out is also what leaves linking open later: linking a second method becomes pointing another provider identity at an existing subject id, authorized by an existing session the way `siwe/link` already works. **No linking flow is built here**, only left unforeclosed.

This is deliberately not v1's shape. v1 created a `users` row per provider identity with a placeholder `publicKey` (`pending-core-kit-<id>`) and used the user id as `verifierId`; that breaks the invariant that an account IS the derived key, and leaves junk accounts behind from abandoned login attempts.

Resolution is insert-then-read against the unique `(kind, identifier_hash)` index rather than an unguarded check-then-insert, so concurrent first logins for one identity yield one subject. Identifiers are stored SHA-256 hashed, following the `auth_methods` convention. **No identity endpoint creates a `users` row.**

Methods do not cross-link: signing in with Google and with email as the same person yields two accounts. That matches v1's documented behavior and is intentional.

## Notes for review

- **`/auth/siwe/login` and `/auth/siwe/link` are untouched** — the engine's Rust client and the contract suite call them. Under D2 `siwe/login` becomes largely redundant, since a wallet can now reach its vault as a first login rather than only authenticating an already-linked one. Worth deciding separately whether it survives; removing it here would be a contract break outside this PR's scope.
- **The identity endpoints are not in the Rust contract suite.** The exchange happens before a login secret exists, so the TypeScript host calls them directly — the ADR explicitly rejects an unstarted engine command surface. They are covered by `apps/api/src/auth/identity.http.itest.ts` against real Postgres instead.
- **The JWKS caching hazard from v1 is handled.** The signing key is required in every deployed profile, allowlisted exactly as `buildJwtOptions` allowlists the access-token secret — v1 only required it in `production`, so staging hit it. The public JWK is derived from the public key rather than by stripping private fields off the private JWK, so no private field can reach the JWKS by omission.
- **Google's audience check is enforced in every profile.** An unconfigured dev profile gets a placeholder audience that no real token can match, so it fails closed rather than warn-and-skipping as v1 did.
- **The token carries no email claim.** The `verifierId` is an opaque uuid, so Torus cannot link a subject to an address. The cost is that a page reload restores the Core Kit session but not the display email, so `UserMenu` shows `[an0n]` until the next explicit login. The login *method* survives a reload, read off the token's own `method` claim. Reversible by adding an `email` claim if the display matters more than the privacy — a deliberate trade, not an oversight.
- **New deployment configuration**, documented in `apps/api/.env.example`: `IDENTITY_JWT_PRIVATE_KEY` (base64 PKCS8 PEM — multiline PEM does not survive a `.env`), `GOOGLE_CLIENT_ID` (the provider's, not the Web3Auth project's), `MAIL_PROVIDER` plus `SENDGRID_API_KEY` and `MAIL_FROM_ADDRESS`. The API refuses to boot in a deployed profile without the signing key or a mail provider. The Web3Auth dashboard also needs a custom verifier pointed at `/auth/.well-known/jwks.json` with `sub` as the verifier ID field.
- `blueprint/api.md` gains the `identity_subjects` row in its data-model list, since that file is normative.

## Verification

Ran green: `pnpm --filter @cipherbox/api test`, `pnpm --filter @cipherbox/web test` (`tsc -b` first), `pnpm lint`, `pnpm lint:md`, `pnpm lint:tracker-refs`, `pnpm --filter @cipherbox/api typecheck`, and `openapi:generate` (committed).

`test:integration` could not run locally — no Docker or Postgres in this environment, so all integration files fail on `ECONNREFUSED 127.0.0.1:5432`, pre-existing ones included. CI runs them.

Tests assert behavior, not source text. Among them: a token signed by a different key is refused; an expired code is refused, as is one CipherBox never issued; a code is single-use and attempt-capped; the JWKS serves no `d`/`p`/`q`/`dp`/`dq`/`qi`; a wallet signature mints a `sub` stable across repeat logins; the same identity logging in concurrently yields one subject row; identity endpoints leave `users` empty; and boot fails in a deployed profile without the signing key, without a mail provider, and without a Google client ID.

## Needs human verification

No dev server or configured verifier exists in the build environment, so Puppeteer verification was not attempted. Against a real Web3Auth custom verifier and a real Google client:

1. Google, email and wallet each sign in from a cold session and reach the same vault for the same provider identity.
2. The same wallet reaches the same account on a second login, and a second wallet reaches a different one.
3. A real code arrives by email from CipherBox and signs in; an expired one is refused.
4. The Core Kit accepts the token against the verifier's JWKS fetch — the one step no local test can exercise.
5. An API restart with `IDENTITY_JWT_PRIVATE_KEY` set does **not** break the next login, which is the v1 failure this guards against.

> [!NOTE]
> ### Issue CipherBox identity tokens from the API and treat wallet login as a first-login flow
> - Adds `IdentityController` with endpoints for Google ID token exchange, email OTP send/verify, SIWE wallet signature, and a JWKS discovery route; all return short-lived RS256 identity tokens signed by `IdentityTokenService`.
> - Introduces `EmailOtpService` (6-digit uniform codes, single-use, attempt cap, per-address send rate limit) and `GoogleOAuthService` (jose-backed RS256 ID token verification) as the credential verification layer.
> - Adds `IdentitySubjectService` to resolve or create a stable `identity_subjects` row per (kind, identifierHash), with concurrency-safe insert-on-conflict; backed by a new DB migration.
> - Replaces the Web3Auth OAuth login path with `loginWithJWT` (verifier + verifierId + idToken); wallet login now performs a full identity-exchange round-trip via the API instead of the previous engine-facilitated SIWE path.
> - Splits the `EmailLoginForm` into a two-step send-then-verify flow; adds `GoogleLoginButton` backed by Google Identity Services (GIS), loaded once per document.
> - Risk: `useAuth` now requires an `IdentityProvider` ancestor with an exchange set, and `loginWithGoogle` now requires a Google ID token argument — both are breaking interface changes for callers.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 164f1c6.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added Google, email verification-code, and wallet authentication options.
  * Added secure, single-use six-digit email codes with expiration, rate limits, and delivery support.
  * Added identity-token exchange and consistent session support across authentication methods.
  * Added Google Identity Services login integration with clearer loading and unavailable states.
* **Documentation**
  * Updated setup guides and environment templates with Google OAuth, email delivery, and identity authentication requirements.
* **Tests**
  * Expanded coverage for authentication flows, validation, security checks, error handling, and end-to-end scenarios.


FSM1:
@coderabbitai review

---
_Generated by [Claude Code](https://claude.ai/code)_

FSM1:
@coderabbitai review

---
_Generated by [Claude Code](https://claude.ai/code)_

FSM1:
@coderabbitai full review please

FSM1:
All 21 review items dispositioned. Head is now `b9202ae`, rebased onto `a7cf06f`.

## Fixed (15)

Every finding verified against the code before acting; three were genuine user-facing defects.

**Majors**

- **A missing `VITE_GOOGLE_CLIENT_ID` disabled every login method, not only Google.** Correct, and the sharpest catch here. `SESSION_ENV` (the two Web3Auth values) is now split from `LOGIN_ENV`, so `loginEnv` no longer requires or returns the Google client ID while `DEPLOY_ENV` and the build gate stay exactly as they were. `coreKit.ts` needed no change — it already destructured only the two values it uses, so the throw *was* the entire bug.
- **The Google button did not survive a busy cycle.** Correct, though the stated mechanism was slightly off in a way that matters for the test: React reuses the same `div` node across the ternary, so node identity survives and a `toBe(target)` assertion passes against the buggy code. What is actually lost is the node's *children* — GIS's injected markup — which the `[clientId]`-only effect never re-renders. The test now injects real button markup and asserts that survives, and it fails against the old code.
- **The staging CSP blocked Google Identity Services.** Correct. `script-src` now allows `https://accounts.google.com/gsi/client`. `frame-src` was deliberately left undeclared: with no `default-src`, GIS's iframe already loads, and adding the directive would be a narrowing that risks breaking WalletConnect.

**Others**: `identity_subjects` added to the complete data-model list; `ci` documented for `VITE_ENVIRONMENT` (and two neighbouring rows corrected that the first fix falsified); test-fake credential fidelity; focus moved to the code field after a successful send; JWKS verification through `createLocalJWKSet`; `afterEach` stub restore; explicit `NODE_ENV` and broader provider assertions in the module-graph test; attempt-cap boundary messages; the named-constant read; `useLayoutEffect` for the GIS callback ref; a refused-verification case; and `forbidNonWhitelisted` coverage.

The JWKS one is worth calling out: the suite verified with `importJWK(keys[0])` and an explicit algorithm, so it never exercised `kid` selection. Signing with a `kid` the JWKS does not publish passed 17/17 before and fails with `JWKSNoMatchingKey` now — a real gap between the tests and how Web3Auth actually verifies.

## Not changed (6), with reasons

- **Email on session restore** — deliberate and already in the PR body. The token carries no email claim precisely so Torus cannot link a subject to an address; the cost is `[an0n]` in `UserMenu` until the next explicit login. The login *method* does survive a reload, off the token's own claim.
- **In-memory OTP state** — matches the existing `ChallengeService` precedent for the SIWE nonce store. Changing one and not the other would be worse than keeping both consistent; they should move together if they move.
- **`kid` rotation** — real, but this PR introduces the signing key rather than managing its lifecycle. Worth its own issue. Verification now goes through `createLocalJWKSet` rather than an index-picked key, which is the precondition for testing any rotation scheme later.
- **Making `API Integration (real Postgres)` a required check** — branch-protection configuration, not something this diff can set.
- **`signature: string`** — the `0x`-prefixed hex is the wire form wagmi returns and the API consumes verbatim; converting to `Uint8Array` here would decode and re-encode a value neither end treats as bytes. Documented on the declaration instead.
- **Docstring coverage 55.32%** — CodeRabbit's default threshold, not a gate this repo sets.

## On the inconclusive linked-issues check

The check asks for evidence that desktop hides wallet login. There is no desktop frontend to hide it in — `apps/desktop` is a Tauri skeleton with no TypeScript sources and no login surface at all. Wallet being web-only is currently enforced by that absence; desktop's own login surface, and the evidence for it, belong to the issue that builds it.

## Verification

Web 327/327, API 227/227, integration 153/153 against real Postgres 16, typecheck, all three lint gates, `openapi:generate` with no diff — all re-run after the rebase. The first fix was also proved in a browser: with `VITE_GOOGLE_CLIENT_ID` unset and the Web3Auth values set, the page renders with email and wallet available and Google shown unavailable, zero page errors; a `staging` build without it still refuses, so the deploy gate is intact.

One pre-existing item noticed while fixing the data-model list: it also omits `device_approvals`, which the section above names. Left alone as outside this diff.

---
_Generated by [Claude Code](https://claude.ai/code)_

